PR #15, opened 2026-03-25
Add lazy_with_connector + Unix socket support

iantay-ant:
Adds `Http2Connection::lazy_with_connector` and `connect_with_connector` as the generic transport escape hatch (same pattern as tonic's `Endpoint::connect_with_connector`). Callers provide a `tower::Service<Uri>` that yields any `hyper::rt::Read + Write` stream; the library boxes it internally and runs the h2 handshake.

`lazy_unix` / `connect_unix` are thin convenience wrappers demonstrating the pattern — they build a `service_fn` closure that dials `tokio::net::UnixStream`.

`BoxedIo` is un-gated from `client-tls` since the custom-connector path uses the same type erasure. The box allocation is once per connection establishment, not per request.

## API

```rust
// Generic — any transport
pub fn lazy_with_connector<C>(connector: C, authority: Uri) -> Self
where
    C: tower::Service<Uri> + Send + 'static,
    C::Response: hyper::rt::Read + hyper::rt::Write + Send + Unpin + 'static,
    C::Error: Into<BoxError>,
    C::Future: Send + 'static,

// Convenience — Unix domain socket
#[cfg(unix)]
pub fn lazy_unix(path: impl Into<PathBuf>, authority: Uri) -> Self
```

## Example

```rust
let conn = Http2Connection::lazy_with_connector(
    tower::service_fn(|_uri: Uri| async {
        let stream = tokio::net::UnixStream::connect("/tmp/app.sock").await?;
        Ok::<_, std::io::Error>(hyper_util::rt::TokioIo::new(stream))
    }),
    "http://localhost".parse().unwrap(),
);
```

### DIFF
--- a/connectrpc/src/client/http2.rs
+++ b/connectrpc/src/client/http2.rs
@@ -311,7 +311,7 @@ impl Http2Connection {
     ///
     /// # Example
     ///
-    /// ```no_run
+    /// ```ignore
     /// # use connectrpc::client::Http2Connection;
     /// # use http::Uri;
     /// let conn = Http2Connection::lazy_with_connector(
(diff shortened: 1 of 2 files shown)


iantay-ant:

iainmcgin (approved):

PR #16, opened 2026-03-26
fix: use service name const instead of string literal in generated clients

### LINKED ISSUE #9: Generated client repeats service name literal instead of using the const
Codegen emits a `*_SERVICE_NAME` const at the top of each generated
service file:

```rust
pub const ELIZA_SERVICE_SERVICE_NAME: &str = "connectrpc.eliza.v1.ElizaService";
```

The router registration path uses it (3 sites in the ELIZA example). But
the generated client methods repeat the string literal:

```rust
pub async fn say_with_options(&self, request: ..., options: CallOptions)
    -> Result<..., ConnectError>
{
    call_unary(
        &self.transport,
        &self.config,
        "connectrpc.eliza.v1.ElizaService",  // <-- could be ELIZA_SERVICE_SERVICE_NAME
        "Say",
        request,
        options,
    ).await
}
```

Three call sites per service (one per RPC in the `_with_options` methods).

The dispatcher's `strip_prefix("connectrpc.eliza.v1.ElizaService/")`
calls use a different form (trailing slash) so can't directly reference
the const - those stay as-is.

## Fix

One-line change in `connectrpc-codegen/src/codegen.rs` to emit the const
ident instead of the literal at client call sites. No behavior change,
just code hygiene.

paikend:
## Summary

Addresses #9 — the codegen emits a `*_SERVICE_NAME` const for each service, but generated client methods were repeating the fully-qualified name as a string literal instead of referencing the const.

- Generated client methods (`call_unary`, `call_server_stream`, `call_client_stream`, `call_bidi_stream`) were using the fully-qualified service name as a **string literal**, even though the codegen already emits a `*_SERVICE_NAME` const at the top of each generated file.
- The server-side router already references the const (3 sites per service in the ELIZA example), but client call sites did not — this inconsistency meant renaming a service required updating both the const definition and every client call site.
- The dispatcher's `strip_prefix("service.name/")` calls use a trailing-slash form and stay as-is (as noted in the issue).

## Problem (from #9)

```rust
// Codegen emits this const:
pub const ELIZA_SERVICE_SERVICE_NAME: &str = "connectrpc.eliza.v1.ElizaService";

// But generated client methods repeated the literal:
call_unary(
    &self.transport, &self.config,
    "connectrpc.eliza.v1.ElizaService",  // ← should be ELIZA_SERVICE_SERVICE_NAME
    "Say",
    request, options,
).await
```

## Fix

Changed `generate_client_method` to accept the const `Ident` and emit it at all four call sites:

```rust
// AFTER:
call_unary(
    &self.transport, &self.config,
    ELIZA_SERVICE_SERVICE_NAME,  // ← references the const
    "Say",
    request, options,
).await
```

## Changes

- `connectrpc-codegen/src/codegen.rs`:
  - Added `Ident` to `proc_macro2` import
  - `generate_client_method` now takes `service_name_const: &Ident` + `full_service_name: &str` (kept for doc strings)
  - All 4 call sites (`call_unary`, `call_server_stream`, `call_client_stream`, `call_bidi_stream`) emit the const ident instead of the string literal

## Type safety

- `call_unary` etc. expect `service: &str`
- The const is `pub const *_SERVICE_NAME: &str = "..."` → coerces to `&str` ✓

## Test plan

- [x] `cargo check --package connectrpc-codegen` — compiles cleanly
- [x] `cargo test --package connectrpc-codegen` — 10 tests passed
- [ ] Regenerate examples with `buf generate` to verify generated output uses the const (requires buf CLI + protoc)

Closes #9


paikend:

iainmcgin:
Thanks for the fix! I'll push the formatting fix onto your branch and merge.

iainmcgin (approved):
LGTM. Service-name const reference brings client emission into structural symmetry with the server-side dispatcher (which already uses the same const at 4 sites). Pushed an rustfmt fix to wrap the call.

PR #17, opened 2026-03-27
[benches] Update RPC benches for current client API

washanhanzi:
## Summary
This updates the RPC benchmark code to match the current connectrpc client API. The client-stream benchmarks now call `client_stream(messages)` without redundant default call options, the manual bidi benchmark request uses `connectrpc::client::full_body(...)`, and the stale `CallOptions` import was removed.

## Why
The benchmark code had drifted behind the current client API and helper usage.

## Impact
The RPC benchmark crate builds cleanly against the current API, keeping the benchmark suite aligned with the library surface.

## Root cause
The benchmark code still referenced an older client-stream call shape and body-construction path after the client helpers changed.

## Validation
- `cargo check --manifest-path benches/rpc/Cargo.toml --benches`

### DIFF
--- a/benches/rpc/benches/cross_impl_bench.rs
+++ b/benches/rpc/benches/cross_impl_bench.rs
@@ -6,7 +6,7 @@ use std::time::Duration;
 use criterion::{BenchmarkId, Criterion, Throughput, criterion_group, criterion_main};
 
 use connectrpc::Protocol;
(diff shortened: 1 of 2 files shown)


washanhanzi:

washanhanzi:
recheck

iainmcgin (approved):
Thanks for the contribution! In #25 I'll enable `--all-targets` to avoid this regressing again in future.

PR #32, opened 2026-04-02
fix(codegen): use fully qualified paths to prevent E0252 on multi-service include

windsornguyen:
When `protoc-gen-buffa-packaging` generates a flat `mod.rs` that `include!`s multiple service files into one Rust module, each file's identical top-level `use` statements collide:

```
error[E0252]: the name `Future` is defined multiple times
 --> generated/connect/svc.v1.machine.rs:1:5
```

This happens in any proto package with more than one service file, because `generate_connect_services` emits the same import block (`use std::future::Future`, `use std::pin::Pin`, `use ::connectrpc::*`, etc.) at the top of every generated file.

## Summary

Replace all top-level `use` statements in generated code with fully qualified paths. The `let imports = quote! { ... }; tokens.extend(imports);` block in `generate_connect_services` is removed entirely. Every reference to the previously imported names now uses its absolute path inline (e.g., `Arc` becomes `::std::sync::Arc`, `Router` becomes `::connectrpc::Router`).

The generated output is functionally identical. Same traits, same impls, same types. No public API changes.

## Implementation

Single file change in `connectrpc-codegen/src/codegen.rs`:

- Deleted the `imports` block from `generate_connect_services` (lines 415-429)
- Updated every `quote!` block that referenced the imported names to use fully qualified paths:
  - `Arc` -> `::std::sync::Arc`
  - `Future` / `Pin` -> `::std::future::Future` / `::std::pin::Pin`
  - `Context`, `ConnectError`, `Router`, `Dispatcher`, handler fns -> `::connectrpc::` prefixed
  - `ClientConfig`, `ClientTransport`, `CallOptions`, call fns -> `::connectrpc::client::` prefixed
  - `Stream` -> `::futures::Stream`
  - `Message` / `OwnedView` -> `::buffa::Message` / `::buffa::view::OwnedView`
  - `__Bytes` / `__CodecFormat` / `__crpc_codegen` aliases -> inlined as `::buffa::bytes::Bytes`, `::connectrpc::CodecFormat`, `::connectrpc::dispatcher::codegen`
- Updated the `compile_precompiled_descriptor_set` test in `connectrpc-build` to match the new output format

## Tests

- `no_top_level_use_statements_in_generated_code` -- verifies generated output contains no `use std::` or `use ::connectrpc` lines
- `multi_service_include_no_e0252` -- generates code for two service files and verifies both can be included in the same module without name collisions
- All 309 existing tests pass. `cargo check --workspace` and `cargo test --workspace` clean.

## Motivation

Any proto package with multiple service definitions triggers E0252 when using `buffa-packaging`'s default `strategy: all` output. The workaround is a manually maintained `mod.rs` that wraps each `include!` in a private submodule and re-exports via `pub use`. This is fragile and gets overwritten on every `buf generate` run.

### DIFF
--- a/connectrpc-build/src/lib.rs
+++ b/connectrpc-build/src/lib.rs
@@ -598,10 +598,16 @@ mod tests {
         // Service trait + client from connectrpc-codegen.
         assert!(content.contains("pub trait EchoService"));
         assert!(content.contains("pub struct EchoServiceClient"));
-        // Absolute crate path (the module-collision fix).
+        // Fully qualified paths (the module-collision fix): no top-level `use`
+        // statements, all references are inline absolute paths like
+        // `::connectrpc::Context`, `::std::sync::Arc`, etc.
         assert!(
-            content.contains("use ::connectrpc::"),
-            "service imports should use ::connectrpc:: absolute path"
+            content.contains("::connectrpc::"),
+            "service code should use ::connectrpc:: fully qualified paths"
+        );
+        assert!(
+            !content.contains("\nuse "),
+            "service code should not emit top-level use statements"
         );
 
         // Include file nests under test.echo.v1. Because out_dir() was set
--- a/connectrpc-codegen/src/codegen.rs
+++ b/connectrpc-codegen/src/codegen.rs
@@ -425,24 +425,10 @@ fn generate_connect_services(
 ) -> Result<TokenStream> {
     let mut tokens = TokenStream::new();
 
-    // All crate-level imports use `::connectrpc` (absolute path) so that
-    // proto packages named `connectrpc.*` (e.g. `connectrpc.conformance.v1`)
-    // don't shadow the crate in generated module scopes.
-    let imports = quote! {
-        use std::future::Future;
-        use std::pin::Pin;
-        use std::sync::Arc;
-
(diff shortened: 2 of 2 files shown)


windsornguyen:

windsornguyen:
cc @iainmcgin -- I would greatly appreciate a review when you get a chance! This fixes a compilation error that affects any proto package with multiple service definitions when using `buffa-packaging`'s default flat `include!` output. Currently seeing in production at Dedalus.

iainmcgin:
Thanks for the contribution! I had been trying to use imports to help keep the generated code a little more readable, but it looks like this has come with unfortunate side-effects. I'm reviewing right now, expecting to approve or fix-forward on your branch imminently.

iainmcgin (approved):
[claude code] Pushed two follow-up fixes (commit 43b2686) and rebased on main:

1. Qualified bare `http_body::Body` to `::http_body::Body` for consistency with all other crate references in the generated output.
2. Fixed broken doc link `[`ClientConfig`]` → `[`connectrpc::client::ClientConfig`]` (this was breaking the Documentation CI job — when the original `use` was removed, rustdoc could no longer resolve the bare reference).
3. Replaced the `!formatted.contains("\\nuse ")` regression check with a syn-based assertion that walks the parsed file's top-level `Item::Use` items, so the test won't false-positive on doc comments or string literals containing the substring.

Rebase resolved a non-semantic conflict in the test module against PR #35 (`emit_register_fn`); both sets of new tests are kept. CI green across all 11 jobs. Thanks for the contribution — nice catch on the workspace tokio audit and the multi-service include scenario.

PR #37, opened 2026-04-02
Update outdated dependencies

iainmcgin:
## Summary

Bumps direct dependencies that were several majors behind (per [deps.rs](https://deps.rs/repo/github/anthropics/connect-rust)) and pins direct deps to security-advisory floors while preserving each crate's major caret. `cargo update` refreshes the rest of the transitives.

## Direct deps bumped (major)

| Dep | Old | New | Notes |
|---|---|---|---|
| `rcgen` | 0.13 | 0.14 | API migration in two test files (see below) |
| `webpki-roots` | 0.26 | 1 | No code changes — `TLS_SERVER_ROOTS` API surface unchanged |
| `pprof` | 0.14 | 0.15 | No code changes |
| `rand` | 0.9 | 0.10 | `Rng` trait split into `Rng` + `RngExt` in 0.10 — `random_range` moved to `RngExt`, so the import changes accordingly |

### rcgen 0.14 migration

`CertificateParams::signed_by` no longer takes `(public_key, ca_cert, ca_key)` — instead it takes `(public_key, &Issuer)`. The new `CertifiedIssuer::self_signed(params, key)` constructor returns a value that holds both the self-signed cert (for the trust store) and the signing identity (via Deref-to-Issuer for issuing leaves), so it's a clean drop-in for the "CA that needs to both sign leaves and be added to a root store" pattern in `peer_certs_reach_handler`.

The `CertifiedKey::key_pair` field was also renamed to `signing_key` (used in tests/streaming).

## Security-advisory pins

Pinned to the patched floor while keeping the original major caret semantics:

| Dep | Old | New | Advisory |
|---|---|---|---|
| `tracing-subscriber` | 0.3 | 0.3.20 | [RUSTSEC-2025-0055](https://rustsec.org/advisories/RUSTSEC-2025-0055) — ANSI escape sequence injection in logged user input |
| `regex` | 1 | 1.5.5 | [RUSTSEC-2022-0013](https://rustsec.org/advisories/RUSTSEC-2022-0013) — DoS via complex regex parsing |

`0.3.20` is still a caret pin (`>=0.3.20, <0.4.0`); same shape as before, just with a higher floor. Same for `1.5.5`. Touched in conformance, eliza, and multiservice for `tracing-subscriber`; eliza only for `regex`.

## Held back

`criterion` stays at 0.5 — `pprof 0.15` still pins `criterion ^0.5` for its `PProfProfiler` bench integration. Bumping criterion past 0.5 would mean giving up the bench-time profiler hook. Worth revisiting once pprof loosens that constraint.

## Transitive updates (selected)

`cargo update` picked up many compatible bumps: `tokio` 1.49 → 1.50, `hyper` 1.8 → 1.9, `zerocopy` 0.8.40 → 0.8.48, `tempfile` 3.26 → 3.27, plus the usual icu/wasm-bindgen/symbolic refreshes. `Cargo.lock` is gitignored (library workspace) so the lockfile diff isn't in this PR, but these will all be picked up on the next `cargo build` against the new `Cargo.toml` constraints.

## Test Plan

- `task fmt`, `task clippy --workspace --all-features -- -D warnings`, `task test --workspace --all-features` — all clean.
- All six conformance suites at the expected counts:
  - Server default: **3600/3600**
  - Server connect-only: **1192/1192**
  - Server connect-tls: **2396/2396**
  - Client Connect: **2580/2580**
  - Client gRPC: **1454/1454**
  - Client gRPC-Web: **2838/2838**


tests/streaming/Cargo.toml
```diff
@@ -25,7 +25,7 @@ connectrpc-build = { path = "../../connectrpc-build" }
 
 [dev-dependencies]
 tokio = { workspace = true, features = ["full", "test-util"] }
-rcgen = "0.13"
```
rpb-ant, line 28: dunno why this isn't `workspace = true`

examples/multiservice/Cargo.toml
```diff
@@ -37,7 +37,7 @@ axum = { workspace = true }
 
 # Logging
 tracing = { workspace = true }
-tracing-subscriber = { version = "0.3", features = ["env-filter"] }
+tracing-subscriber = { version = "0.3.20", features = ["env-filter"] }
```
rpb-ant: dunno why this isn't `workspace = true`

rpb-ant (dismissed):

tests/streaming/Cargo.toml
iainmcgin: an oversight - will fix this (and examples/multiservice). Some crates are specifically isolated from the root workspace, but these aren't.

rpb-ant (approved):

PR #41, opened 2026-04-03
Prepare connectrpc 0.3.2 release

iainmcgin:
Lockstep bump of all three crates to 0.3.2: `connectrpc`, `connectrpc-codegen`, `connectrpc-build`.

## Headline

#32 — generated service code no longer emits top-level `use` statements, so multiple service files can be `include!`d into the same Rust module without E0252 collisions. This was the proximate motivation for cutting 0.3.2 (it unblocks downstream consumers using buffa-packaging's flat output strategy).

## Other changes shipping in 0.3.2

- **#16** — generated client methods reference the per-service `*_SERVICE_NAME` const instead of repeating the fully-qualified service name as a string literal at every call site.
- **#19** — `wasm32-unknown-unknown` target compatibility for the `connectrpc` crate (default features off), plus a new `examples/wasm-client` demonstrating a Fetch-based `ClientTransport`. As a side effect of this PR, the workspace `tokio` feature footprint narrowed: published `connectrpc` previously inherited the full workspace tokio feature set when `workspace = true` was inlined at publish time, and now requests only `rt`, `io-util`, `sync`, `time` (plus `net` when the `client`/`server` feature is enabled). Downstream crates that use `tokio` directly should declare their own features rather than relying on transitive activation.
- **#37** — workspace dependency updates.

## Notes

- Path-dep version constraint (build → codegen) already uses caret `"0.3"`, so no change required there.
- README dependency snippets are caret `"0.3"`, unchanged.
- CHANGELOG gains a 0.3.2 section with the four entries above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,49 @@ increment the patch version.
 
 ## [Unreleased]
 
+## [0.3.2] - 2026-04-03
+
+### Fixed
+
+- **Generated service code now compiles when multiple services are
+  `include!`d into the same Rust module** ([#32]). The codegen previously
+  emitted top-level `use` statements that collided with E0252 when
+  buffa-packaging's flat-output strategy concatenated several service
+  files into one module. Bindings now use fully-qualified paths
+  throughout (`::connectrpc::Context`, `::buffa::view::OwnedView`,
+  `::http_body::Body`, etc.), so multiple service files can coexist in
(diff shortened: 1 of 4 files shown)


iainmcgin:
cc @windsornguyen once this is pushed to crates.io, let me know if it solves your issues or if there is more to do!

aaomidi (approved):

PR #55, opened 2026-04-17
fix: do not require default features for axum

siketyan:
Previously, the connectrpc crate required default features of the `axum` crate. This created a dependency chain of `axum/tokio` -> `tokio/net` -> `mio`. Since `mio` does not support WebAssembly (WASM), it was impossible to use this crate in environments like Cloudflare Workers.

Because `connectrpc` does not actually depend on any of the default features, this PR disables them. This change removes the incompatible dependencies and enables the crate to be used in WASM environments.

No changes are required to the current crate users.


siketyan:

rpb-ant (approved):

iainmcgin:
Thanks for the contribution!

PR #56, opened 2026-04-17
connectrpc-build: skip rerun-if-changed for .files() in Precompiled mode

sharma-uday:
In Precompiled mode (`Config::descriptor_set(path)`), `.files()` entries are proto-relative names per the docs, not on-disk paths. Emitting `cargo:rerun-if-changed` for them points cargo at nonexistent files, which it treats as permanently stale — the consumer crate rebuilds on every `cargo build` invocation.

The `.pb` path is the only real input in Precompiled mode and already has its own `rerun-if-changed` line. Protoc/Buf modes are unchanged.

**Repro:** any crate using `.descriptor_set("foo.pb").files(&["pkg/foo.proto"])` where `pkg/foo.proto` doesn't exist relative to the crate dir — `CARGO_LOG=cargo::core::compiler::fingerprint=info cargo build` shows `StaleItem(MissingFile { ... })` on every run.

**Testing:** `cargo build/clippy/fmt/test -p connectrpc-build` clean. Verified against a local Precompiled-mode consumer: 4× consecutive `cargo build` are now no-ops (were full rebuilds before).

### DIFF
--- a/connectrpc-build/src/lib.rs
+++ b/connectrpc-build/src/lib.rs
@@ -274,12 +274,20 @@ impl Config {
             write_if_changed(&include_path, include_src.as_bytes())?;
         }
 
-        // 6. Cargo re-run triggers.
-        for f in &self.files {
-            println!("cargo:rerun-if-changed={}", f.display());
-        }
-        if let DescriptorSource::Precompiled(p) = &self.descriptor_source {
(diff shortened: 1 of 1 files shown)


sharma-uday:

iainmcgin (approved):

PR #58, opened 2026-04-20
Bring graceful shutdown in line with Go's net/http.Server.Shutdown behavior

rpb-ant:
connect-rust doesn't currently handle graceful shutdown very gracefully. I'm mostly familiar with Go, where the http stack has some useful conventions:
- stop accepting new tcp connections
- allow in-flight requests (or h2 streams) to complete

For h1 it tears the connection down as soon as it goes idle (and sets `Connection: close` on the in-flight response if there is one). For h2, since streams can overlap, it uses the GOAWAY frame — actually a kind of clever GOAWAY → ping → pong → GOAWAY(last_stream_id) sequence so streams that race the first GOAWAY are still handled.

hyper implements the same behaviors, so this PR is mostly just plugging that in. Mechanically: each accepted connection gets wrapped with `hyper_util::server::graceful::Watcher::watch(conn)`, and after the accept loop exits we call `graceful.shutdown().await` to signal + drain. This fully subsumes the old `TaskTracker` (the `Watcher` is created in the accept loop and moved into the per-conn task, so `shutdown()` waits for tasks still mid-TLS-handshake too), so that's removed along with the `tokio-util/rt` feature.

Tests: added an `h2` dev-dep and a regression test that opens a raw h2 connection and asserts the server actually GOAWAYs it on shutdown. Also rewrote the old `with_inflight_connection` test — it was asserting the previous "idle socket keeps the server alive" behavior, which is exactly what we're fixing — into `drains_inflight_request`, which checks that an actual in-flight request is allowed to finish.

The doc comment that previously called this out as a Limitation now describes the new shutdown sequence instead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,7 @@ tracing = "0.1"
 tracing-subscriber = { version = "0.3.20", features = ["env-filter"] }
 
 # HTTP server/client
+h2 = "0.4"
 hyper = "1"
 hyper-util = "0"
 
--- a/connectrpc/Cargo.toml
+++ b/connectrpc/Cargo.toml
@@ -102,7 +102,6 @@ server = [
     "dep:libc",
     "dep:tower-http",
     "tokio/net",
-    "tokio-util/rt",
 ]
 
 # HTTP client with connection pooling
--- a/connectrpc/Cargo.toml
+++ b/connectrpc/Cargo.toml
@@ -76,6 +76,7 @@ tokio = { workspace = true, features = ["macros", "rt-multi-thread", "test-util"
 tempfile = { workspace = true }
 rcgen = { workspace = true }
 buffa-types = { workspace = true }
+h2 = { workspace = true }
 
 [features]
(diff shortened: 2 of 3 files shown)


rpb-ant:

iainmcgin (approved):
It may be worth considering for a follow-up whether to offer [multiple phases of shutdown grace period](https://doc.traefik.io/traefik/v1.7/configuration/commons/#life-cycle): this can help compensate for slow updates on valid targets in load balancers, and propagation delays in K8s service resources / DNS records.

rpb-ant:
I like multi-phase graceful shutdown, but in my experience the cleanest approach is for the _application itself_ to handle the first phase: broadcasting its drain status to clients (including loadbalancers) with whatever health-checking protocol it happens to employ.

Once it has given its own drain status time to propagate, _that's_ when it should initiate the graceful shutdown at the connect-rust layer (which should handle closing the listening, no longer accepting new connections, doing the awkward GOAWAY --> ping --> ping --> GOAWAY dance)

PR #63, opened 2026-04-27
service: spawn body reader via wasm-bindgen-futures on wasm32

connyay:
`spawn_body_reader` runs the per-request envelope decoder in the background so bidi and client-streaming RPCs can read incoming messages while the handler is producing responses. It used `tokio::spawn`, which panics with "there is no reactor running, must be called from the context of a Tokio 1.x runtime" on `wasm32-unknown-unknown` targets (Cloudflare Workers, etc.), where the ambient executor is `wasm-bindgen-futures` rather than tokio. Server-streaming and unary RPCs worked on Workers, but bidi and client-streaming panicked the moment the first envelope was decoded.

Under `cfg(target_arch = "wasm32")`, dispatch the reader future via `wasm_bindgen_futures::spawn_local` instead. `spawn_local` doesn't return a joinable handle, but the existing code only held the `JoinHandle` for lifetime association (don't abort early), so `None` works fine. `spawn_body_reader`'s return type and
`StreamingResponseBody::with_reader_task` now take `Option<JoinHandle<()>>`. `tokio::sync::mpsc` keeps working unchanged; its internals only need wakers, which `wasm-bindgen-futures` provides.

Adds a wasm32-only target dep on `wasm-bindgen-futures = "0.4"`. No behaviour change on native targets; the tokio path is unchanged.

Verified end-to-end against a Cloudflare Worker via miniflare and `wrangler dev`: bidi calls now succeed, the handler is dispatched before the request body completes, and per-input echoes round-trip in ~1ms while the request stream is still open. See
https://github.com/connyay/example-connectrpc-worker for a runnable example (clock = server-stream, echo = bidi, heartbeat = full-duplex bidi probe).

Scope here is server-side handler dispatch. `call_bidi_stream` in `client/mod.rs` also calls `tokio::spawn`, but the rest of the `client` module already requires hyper / tokio-net features that don't compile on wasm32. Wasm client transports are a separate concern handled by user-supplied `ClientTransport` impls (see `examples/wasm-client`).


connyay:

iainmcgin:
[claude code] Pushed a small fixup commit (d0ce6b0) extracting a `spawn_detached` helper to centralise the platform fork. The two inline `#[cfg(target_arch = "wasm32")]` blocks in `spawn_body_reader` are replaced with a single `let reader_task = spawn_detached(reader_future);` call, with the cfg gates moved onto two definitions of the helper (one with the `Send + 'static` bound for `tokio::spawn`, one with just `'static` for `spawn_local`). The wasm hedge in `with_reader_task`'s docstring also drops out — the platform contract now lives in one place.

Verified locally:
- `cargo check -p connectrpc` (native) — clean
- `cargo check -p connectrpc --target wasm32-unknown-unknown --no-default-features` — clean
- `cargo test -p connectrpc --lib` — 235/235 pass
- `cargo clippy -p connectrpc --all-targets -- -D warnings` — clean
- `cargo doc --document-private-items` — same warning count as baseline (9 pre-existing intra-doc-link warnings unrelated to this PR)

Happy to revert if you'd rather keep the original shape, but this gets the diff to a place we're comfortable approving and merging.

connyay:
The fixup looks great, thanks!

iainmcgin (approved):

PR #69, opened 2026-04-28
connectrpc-build: add Config::emit_rerun_directives(bool)

### LINKED ISSUE #68: connectrpc-build: allow suppressing cargo:rerun-if-changed for non-Cargo callers
`connectrpc_build::Config::compile()` unconditionally prints
`cargo:rerun-if-changed=...` lines to stdout. That's correct for a Cargo
`build.rs` context, but when the same `Config` is driven from a
non-Cargo host tool (e.g. a Bazel genrule that wraps the codegen as a
binary) those lines are noise on stdout - the wrapping rule has to
capture and discard them to keep build output clean.

Current workaround in the Bazel rule:

```python
# Capture all tool output to a declared sibling and forward it to
# stderr only on non-zero exit, so the noise is hidden on success
# but diagnostics survive on failure.
```

Proposed fix: add `Config::emit_rerun_directives(bool)` (default `true`)
so non-Cargo callers can opt out:

```rust
connectrpc_build::Config::new()
    .files(&protos)
    .out_dir(out)
    .emit_rerun_directives(false)
    .compile()?;
```

iainmcgin:
Fixes #68

`Config::compile()` unconditionally prints `cargo:rerun-if-changed=...`
to stdout. Correct for `build.rs`; noise for non-Cargo callers (Bazel
genrules, standalone host tools) that have to capture and discard it.

Adds `Config::emit_rerun_directives(bool)` (default `true`) so callers
outside a Cargo build script can opt out:

```rust
connectrpc_build::Config::new()
    .files(&protos)
    .out_dir(out)
    .emit_rerun_directives(false)
    .compile()?;
```

Net +29 lines. Additive; default behaviour unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,13 @@ increment the patch version.
   accessors), and the `Debug` output for an unset map now shows
   `None` instead of `{}`.
 
+### Added
+
+- **`connectrpc-build`**: `Config::emit_rerun_directives(bool)` to suppress
+  the `cargo:rerun-if-changed=` lines when running outside a Cargo
+  `build.rs` context (e.g. from a Bazel genrule or standalone host tool).
+  Default remains `true`.
+
 [#34]: https://github.com/anthropics/connect-rust/issues/34
 [#61]: https://github.com/anthropics/connect-rust/issues/61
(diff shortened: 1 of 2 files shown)


zertosh (commented):
❤️

asacamano (approved):

PR #77, opened 2026-05-04
codegen: detect colliding OwnedFooView aliases instead of always emitting

### LINKED ISSUE #75: Name collision when imported proto contains the same message name
## Bug
Consider a file `api/v1/foo/bar/foobar.proto`

```proto
message MyMessage {}
```

Another file `api/v2/foo/bar/foobar.proto`

```proto
import "api/v1/foo/bar/foobar.proto";

service FooBar {
    rpc DoSomething (api.v1.foo.bar.MyMessage) returns (...);
}

message MyMessage {} // used somewhere else
```

Results in compiler error - 
```rust
the name `OwnedMyMessageView` is defined multiple times
```

## Versions
Buffa v0.4.0
connect-rust d0bee62 (git)

iainmcgin:
Fixes #75.

## Problem

`connectrpc-codegen` emits a `pub type OwnedFooView = OwnedView<FooView<'static>>;` shorthand for every RPC input/output type referenced by a service, keyed on the proto FQN's last segment. A service file that defines its own `MyMessage` and also references an imported `.api.v1.foo.bar.MyMessage` produced two `pub type OwnedMyMessageView` declarations in the same module — compile fails with `error[E0428]: the name OwnedMyMessageView is defined multiple times`.

The bug was introduced in v0.4.0 by the handler-ergonomics work (PR #67); reported by @varunbpatil at #75.

## Fix

Detect actual alias-name collisions and suppress the alias only for the colliding set, keeping it for every other type.

A pre-pass over `file_to_generate` collects every service method's input/output FQN, computes the would-be alias name per `(file_package, alias_name)`, and records pairs where two or more distinct FQNs would land at the same identifier in the same target Rust module. The pre-pass needs full-batch visibility because the package-mod stitcher mounts sibling files into one module, so collisions can span files.

For colliding members the trait method signature inlines `::buffa::view::OwnedView<…View<'static>>` directly — the same form the client-side codegen has always used for response types. Non-colliding types (the common case, including same-package types and well-known imports like `.google.protobuf.Empty`) keep their alias unchanged, so existing handler code that wrote `request: OwnedEmptyView` keeps working.

## Tests

Three new tests in `connectrpc-codegen/src/codegen.rs`:

- `cross_package_input_collision_suppresses_alias_for_both_sides` — the #75 reproducer.
- `cross_package_input_without_collision_keeps_alias` — covers the WKT case where a cross-package import has a unique short name; alias must still be emitted.
- `collision_inlines_in_all_streaming_method_shapes` — drives unary, server-streaming, client-streaming, and bidi through a colliding cross-package input. There are four `#input_arg` interpolation sites in `generate_trait_method`; this catches a regression that drops the substitution from any one of them.

All 31 codegen tests pass. Full workspace `cargo check --workspace --all-features --all-targets`, `task lint`, and `task test` clean. The four checked-in `*/generated/*` consumers compile unchanged (none have cross-package RPC inputs, so the codegen output is byte-identical for them).

## Notes

The previous behaviour wasn't strictly preserved — pre-fix, every cross-package input got an alias, and a colliding-set always failed E0428. After the fix, colliding members lose the alias instead of the whole compile failing, and the rest are unchanged. There's no documented user shape that was depending on the failed-to-compile case.

A related but pre-existing pattern is not closed by this PR: same-package nested vs. top-level types with the same short name (e.g. `.foo.bar.Inner` and `.foo.bar.Outer.Inner`) would still collide. Filing as a follow-up issue; the fix is mechanically the same — the pre-pass already operates on `(package, alias_name)` so extending it is a one-line change once we agree on the alias naming policy for nested types.

### DIFF
--- a/connectrpc-codegen/src/codegen.rs
+++ b/connectrpc-codegen/src/codegen.rs
@@ -89,8 +89,13 @@ fn emit_service_files(
     // - output-type Encodable impls (else two files sharing an output
     //   type collide with E0119);
     // - OwnedFooView aliases keyed on (package, fqn) (else two files in
-    //   the same package collide with E0428).
-    let mut batch = BatchState::default();
+    //   the same package collide with E0428);
+    // - colliding-alias detection (issue #75) needs full-batch visibility
+    //   because the stitcher mounts sibling files into one module.
+    let mut batch = BatchState {
+        colliding_aliases: collect_alias_collisions(proto_file, file_to_generate),
+        ..BatchState::default()
+    };
     for file_name in file_to_generate {
         let file_desc = proto_file
             .iter()
@@ -509,6 +514,19 @@ struct BatchState {
     /// `Owned#{Msg}View` alias has already been emitted (per package
     /// module; aliases are module-scoped).
     alias_seen: std::collections::BTreeSet<(String, String)>,
+    /// `(package, alias_name)` pairs where two or more distinct FQNs would
+    /// produce the same `Owned<Msg>View` alias in the same target Rust
+    /// module — e.g. a service file that defines its own `MyMessage` and
+    /// also references an imported `.api.v1.foo.bar.MyMessage` (issue
+    /// [#75]). The alias is suppressed for every member of a colliding
+    /// set; trait method signatures inline the
+    /// `::buffa::view::OwnedView<…<'static>>` form for those types
+    /// instead. Aliases for non-colliding types (the common case,
+    /// including same-package and well-known types like
+    /// `.google.protobuf.Empty`) are unaffected.
+    ///
+    /// [#75]: https://github.com/anthropics/connect-rust/issues/75
+    colliding_aliases: std::collections::BTreeSet<(String, String)>,
 }
 
 fn generate_connect_services(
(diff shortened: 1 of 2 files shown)


rpb-ant (approved):
I'm fine with this approach, but do we want to consider just doing full qualification everywhere, no aliases at all?

iainmcgin:
Thanks for the review. Agreed that uniformity is the right direction — the asymmetry this fix introduces (alias when no collision, inlined form when there is) is the kind of "why does it work for type X but not Y" footgun we'd rather not ship.

Going to land this as-is for 0.4.1 to close #75 for the user who reported it, then plan **uniform inlining + alias removal for 0.5.0** as a deliberate breaking change with a documented migration path.

The broader question this opens up is re-exports across both projects — buffa#80 is highlighting the same ergonomic friction from a different angle (deep `__buffa::oneof::event::Payload` paths are hard to read in code that crosses package boundaries). It's genuinely useful to be able to surface types and aliases at their "expected" location in the tree (`test::EventPayload`-style, or in our case the `OwnedFooView` shorthand at the package level) — the small exception is exactly when that collides with another proto-defined name in the same module. So whatever we settle on for 0.5.0 should think about both: the re-export pattern that gives us the ergonomic shorthand by default, *and* the collision-disambiguation rule that kicks in only when needed. I'll fold this into a 0.5.0 design issue once we work through buffa#80's resolution.

PR #80, opened 2026-05-05
axum: serve_tls helper + mtls-identity example

### LINKED ISSUE #49: examples/mtls-identity: demonstrate cert-SAN identity with axum + a helper for parity with standalone Server
PR #31 added `PeerCerts` and `PeerAddr` to `Context::extensions`, automatically populated by the standalone `Server` from the rustls accept loop. Handlers can read peer identity (cert SANs, remote addr) without caring about transport - on the standalone Server.

For axum users, this is currently a do-it-yourself path: `axum::serve(listener, app)` assumes a plain `TcpListener`, so a custom rustls accept loop is needed to extract `peer_certificates()` and inject it into request extensions per-connection.

This issue tracks two related pieces of work.

## Part 1: `examples/mtls-identity/` example

A new example crate demonstrating mTLS + cert-SAN-based identity extraction in an axum app. Mirrors the existing `examples/middleware/` shape but swaps bearer-token auth for mTLS:

- Server uses an axum router behind a rustls TLS terminator that requires client certificates (`WebPkiClientVerifier`)
- Per-connection layer captures `PeerCerts` and `PeerAddr`, injects them into request extensions
- Handler reads identity by parsing the leaf cert's SANs from `ctx.extensions.get::<PeerCerts>()` rather than from a bearer token
- Test certs generated at runtime via `rcgen` (no PEM files in the repo)
- Integration test exercises authorized + permission-denied paths

The example would show how to bridge axum's listener model with connectrpc's peer-identity convention, so handlers stay transport-agnostic.

## Part 2: Helper API for axum + connectrpc identity parity

Today, the standalone `Server::with_tls` magically populates `PeerCerts`/`PeerAddr` for handlers. Axum users have to write the rustls accept loop and tower-layer wiring themselves. The same handler code that reads `ctx.extensions.get::<PeerCerts>()` won't work on axum without that custom wiring.

Proposal: ship a thin helper from connectrpc that does the rustls accept loop + extension injection for axum users. Sketch:

```rust
// In connectrpc::axum (behind the `axum` + `server-tls` features)
pub async fn serve_tls(
    listener: tokio::net::TcpListener,
    app: axum::Router,
    tls_config: Arc<rustls::ServerConfig>,
) -> Result<(), BoxError> {
    // Custom accept loop:
    //   1. TCP accept
    //   2. tokio_rustls handshake
    //   3. Extract peer_certificates() + peer_addr()
    //   4. Wrap each connection in a hyper service that
    //      injects PeerCerts/PeerAddr into req.extensions before
    //      handing off to the axum service
}
```

With this helper:

```rust
// Plaintext - existing axum convention
axum::serve(listener, app).await?;

// TLS with peer-identity passthrough - drop-in replacement
connectrpc::axum::serve_tls(listener, app, tls_config).await?;
```

The same handler code reads `ctx.extensions.get::<PeerCerts>()` regardless of whether the server was built with the standalone `Server` or axum + this helper. Hosting choice and identity extraction are now orthogonal.

### Alternatives considered

- **Tower layer**: a layer can't extract peer certs from the rustls connection unless something earlier in the stack already exposed them - the layer runs at the HTTP-request level, not the TCP/TLS-connection level. So the helper has to be at the listener/accept-loop level, not a layer.
- **Document the manual pattern in the example only**: doesn't scale - every axum + mTLS user reimplements the same accept loop and per-connection injection layer.
- **Push the work to `axum-server` integration**: adds a transitive dependency for axum users; the accept loop is small enough to ship in connectrpc directly.

## Suggested order

Land Part 2 (the helper) first so the example in Part 1 can use the public API rather than reimplementing it inline. Both can ship in the same release.

## References

- #31 `Expose mTLS peer credentials and remote address to handlers` (the original `PeerCerts`/`PeerAddr` work)
- #46 `Add streaming-tour and middleware examples` (the existing middleware example this would complement)
- `connectrpc/src/server.rs:577` for the existing peer-cert extraction logic in the standalone Server

iainmcgin:
Closes #49.

## What

Two commits, in the order #49 suggests landing them:

**1. `connectrpc::axum::serve_tls`** — a TLS-aware counterpart to `axum::serve(listener, router)` that owns the rustls accept loop and stamps `PeerAddr` / `PeerCerts` into request extensions, the same convention the standalone `Server::with_tls` uses. Handler code that reads `ctx.extensions.get::<PeerCerts>()` is now portable between the standalone `Server` and an axum app.

The returned `ServeTls` mirrors `axum::serve::Serve`'s shape: a `#[must_use]` `IntoFuture<Output = io::Result<()>>` builder with `.tls_handshake_timeout()` and `.with_graceful_shutdown()` knobs. The accept loop reuses the standalone `Server`'s conventions: slowloris-bounded handshake timeout (default `DEFAULT_TLS_HANDSHAKE_TIMEOUT`), transient `accept()` error skip, `TCP_NODELAY`, and hyper-util `GracefulShutdown` drain. Module docs spell out the differences from `axum::serve` (concrete `Router` only, ALPN setup, `PeerCerts` is conditional, no automatic panic catching).

**2. `examples/mtls-identity`** — the mTLS twin of `examples/middleware`: same secret-store-with-ACL shape, but the credential is the client certificate the TLS handshake verified, not a `Bearer` token. The handler reads `PeerCerts` from extensions, parses the leaf cert's DNS SAN with `x509-parser`, derives a single-label workload name under `workloads.example.com`, and gates an ACL on it.

Single self-contained binary: in-memory `rcgen` PKI (CA, server leaf, two workload client leafs), serve, call with both identities, graceful shutdown. No PEM files touch disk. The `lib.rs` exposes the proto, handler, identity extraction, and PKI helpers so `tests/e2e.rs` reuses them rather than duplicating cert generation.

## Side fixes

- **`tokio/macros` added to the `server` feature.** `server.rs` and the new `axum.rs` both use `tokio::select!`, but the macro feature was previously only reachable via dev-dep unification. A downstream crate that enabled `server` without itself enabling `tokio/macros` would fail to compile; CI never caught it because both `--all-targets` (Check) and `cargo test` pull in dev-deps.
- **`docs/guide.md` TLS-hosting drift fix.** The guide claimed eliza wraps axum with `tokio_rustls::TlsAcceptor` for TLS; eliza actually switches to the standalone `Server` for the TLS path. Replaced with the new `serve_tls` snippet.

## Tests

`connectrpc/src/axum.rs` (`#[cfg(test)]`):
- mTLS round-trip: `PeerAddr` + `PeerCerts` reach the handler with the verified leaf
- handshake timeout releases the per-connection watcher so graceful drain completes
- handshake error (garbage bytes) doesn't kill the accept loop; a subsequent valid client succeeds
- graceful shutdown anchors an in-flight request until the handler returns

`examples/mtls-identity` (`tests/e2e.rs` + `lib.rs` unit tests):
- `WhoAmI` reflects cert SAN and remote addr
- authorized read succeeds with `x-served-by` trailer
- permission denied for another workload's secret
- TLS client without a cert is rejected at the handshake
- `extract_identity` rejects no-cert, empty-prefix, multi-label-prefix, and unrelated-domain SANs

## Notes for review

- `pub mod axum` shadows the extern `axum` crate within `connectrpc`'s crate-root scope only. There's an in-source comment explaining the constraint (don't `use axum::...` in `lib.rs`); downstream `use connectrpc::axum::serve_tls;` is unambiguous. I considered `connectrpc::axum_tls` but the issue spec's `connectrpc::axum::serve_tls` reads better, and the shadowing is contained.
- `serve_tls` accepts a concrete `axum::Router`, not the make-service forms `axum::serve` is generic over. The module docs explain why (`PeerAddr` replaces `ConnectInfo<SocketAddr>`, so `into_make_service_with_connect_info` has no analogue). Generalising the bound would be possible but adds API surface I'd rather defer until someone needs it.
- The diff is over the soft 250-line target, split across two commits as #49 suggests. Most of the example's `lib.rs` is the in-memory PKI helper, which is heavily commented because it doubles as a "what a deployment would load from a secret store" model.

## Reviewer findings deferred (Low / Advisory)

Flagging these for a decision rather than addressing them in this PR:

- `ServeTls` doesn't expose an `http1_keep_alive(bool)` knob like `BoundServer` does. `axum::serve` doesn't either, so `serve_tls` is at parity with the thing it replaces. Easy to add later if anyone hits the stale-connection race on axum.
- No accept-rate / max-connections cap on the spawned per-connection tasks. Same shape as the standalone `Server` and `axum::serve`, both unbounded. Could grow a `Semaphore`-backed knob if it becomes a need.
- `PeerAddr` could derive `Copy` (it wraps a `Copy` `SocketAddr`). Public API change in `server.rs`, out of scope here.
- The `axum.rs` test fixture duplicates the `pki()` helper from `server.rs`'s tests. They differ in what they return; hoisting to a shared `#[cfg(test)]` module is possible but increases coupling.

### DIFF
--- a/examples/mtls-identity/src/lib.rs
+++ b/examples/mtls-identity/src/lib.rs
@@ -76,10 +76,9 @@ pub fn extract_identity(certs: Option<&PeerCerts>) -> Result<Identity, ConnectEr
         .flat_map(|ext| ext.value.general_names.iter())
         .find_map(|gn| {
             // Only `<single-label>.workloads.example.com` is a workload SAN.
-            // Reject `.workloads.example.com` (empty label) and
-            // `a.b.workloads.example.com` (would alias as `a.b`), which an
-            // attacker-controlled CA could otherwise mint to spoof an ACL
-            // entry.
+            // Reject `.workloads.example.com` and `a.b.workloads.example.com`:
+            // we intend to accept only direct subdomains of the workload
+            // domain.
             let GeneralName::DNSName(dns) = gn else {
                 return None;
             };
@@ -353,7 +352,7 @@ mod tests {
             extract_identity(Some(&empty)).unwrap_err().code,
             ErrorCode::Unauthenticated
         );
-        // Multi-label prefix: "a.b" would alias an ACL entry; reject it.
+        // Multi-label prefix: not a direct subdomain; reject it.
         let multi = peer_certs_with_dns_sans(&["a.b.workloads.example.com"]);
         assert_eq!(
             extract_identity(Some(&multi)).unwrap_err().code,
--- a/examples/mtls-identity/src/lib.rs
+++ b/examples/mtls-identity/src/lib.rs
@@ -1,0 +1,372 @@
+//! mTLS cert-SAN identity for axum-hosted ConnectRPC services.
+//!
+//! Mirrors `examples/middleware/`, swapping bearer-token auth for mTLS:
+//! instead of an `axum::middleware::from_fn` reading an `Authorization`
+//! header, identity comes from the verified client certificate that
+//! `connectrpc::axum::serve_tls` captures during the TLS handshake and
+//! stamps into request extensions as [`PeerCerts`]. The handler parses
+//! the leaf cert's DNS SAN to derive a workload identity, then enforces
+//! an ACL against it.
+//!
+//! The same handler code works unchanged on the standalone
+//! [`connectrpc::Server::with_tls`], which populates [`PeerCerts`] the
+//! same way — the hosting choice doesn't leak into authorization logic.
+
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use buffa::view::OwnedView;
+use connectrpc::{
+    ConnectError, ErrorCode, PeerAddr, PeerCerts, RequestContext, Router, ServiceResult,
+};
+
+pub mod proto {
+    connectrpc::include_generated!();
+}
+
+pub use proto::anthropic::connectrpc::mtls_identity::v1::*;
+
+pub type BoxError = Box<dyn std::error::Error + Send + Sync>;
+
+// ============================================================================
+// Identity: derive a workload name from the leaf cert's DNS SAN.
+// ============================================================================
+
+/// All clients in this demo carry a SAN under this domain. Anything else
+/// is rejected as `Unauthenticated`.
+pub const WORKLOAD_DOMAIN: &str = "workloads.example.com";
+
+/// Caller identity, parsed from the leaf certificate's DNS SAN.
+#[derive(Debug, Clone)]
+pub struct Identity {
+    /// Short workload name, e.g. `"alice"`.
+    pub name: String,
+    /// Full DNS SAN as presented, e.g. `"alice.workloads.example.com"`.
+    pub san: String,
+}
+
+/// Parse a workload identity out of the leaf certificate's DNS SANs.
+///
+/// Returns the first SAN under [`WORKLOAD_DOMAIN`]. Returns
+/// `Unauthenticated` when no client cert was presented (a non-mTLS
+/// connection) or no SAN matches the expected domain.
+///
+/// In a real deployment you'd typically match a SPIFFE ID
+/// (`spiffe://trust-domain/path`, a URI SAN) instead of a DNS SAN, or
+/// delegate this whole step to an authorization framework. The shape is
+/// the same: read [`PeerCerts`], parse the leaf, derive an identity.
+pub fn extract_identity(certs: Option<&PeerCerts>) -> Result<Identity, ConnectError> {
+    use x509_parser::extensions::GeneralName;
+    use x509_parser::prelude::{FromDer, X509Certificate};
+
+    let leaf = certs.and_then(|c| c.0.first()).ok_or_else(|| {
+        ConnectError::new(ErrorCode::Unauthenticated, "client certificate required")
+    })?;
+
+    let (_, parsed) = X509Certificate::from_der(leaf.as_ref()).map_err(|e| {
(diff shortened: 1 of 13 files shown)


examples/mtls-identity/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+        .ok()
+        .flatten()
+        .into_iter()
+        .flat_map(|ext| ext.value.general_names.iter())
+        .find_map(|gn| {
+            // Only `<single-label>.workloads.example.com` is a workload SAN.
+            // Reject `.workloads.example.com` (empty label) and
+            // `a.b.workloads.example.com` (would alias as `a.b`), which an
+            // attacker-controlled CA could otherwise mint to spoof an ACL
+            // entry.
```
asacamano: I realize this is an example, but I wonder about this...

If an attacker-controlled CA is accepted by the mTLS exchange, could they just spoof everything? Not just "a.b", but also "trustedjob", and also maybe "a.b.workloads.example.com", or "trustedjob.workloads.example.com"?

asacamano (dismissed):

examples/mtls-identity/src/lib.rs
iainmcgin: Yeah, this looks like nonsense, will fix

iainmcgin: [claude code] You're right — if the trust store accepts an attacker-controlled CA, the attacker can mint a cert for *any* SAN, including `trustedjob.workloads.example.com`, so the comment was overstating what this guard buys you. The empty/multi-label checks are about parsing intent (only direct subdomains map to ACL names), not a CA-compromise mitigation. Reworded the comment in 2f5777b to just state the intent and drop the spoofing framing.

asacamano (dismissed):

asacamano (approved):

PR #98, opened 2026-05-12
handler: PreEncoded body type + Encodable items for streaming handlers

rpb-ant:
## Why

A handler that builds a `*View<'a>` from local app state (e.g. a resolver snapshot held in `Arc`) can't satisfy `Handler::Body: 'static` or `ServiceStream<Res>` today — the only ViewEncode-friendly bodies are request-buffer-backed (`OwnedView`, `MaybeBorrowed`). Such handlers need to encode the view internally while the snapshot is still in scope, then return the bytes.

This unblocks the dominant ViewEncode use case for high-throughput servers: 1-source-to-N-readers fanout where each per-reader response borrows from a shared decoded source. Codec-layer benches (`benches/rpc/echo_bloat.rs`, fanout sweep at N=64) show ~3.5× over the move-baseline; resolver-shaped workloads (~1k endpoints with string-map metadata per response) measure 2.8–4.2× depending on compression.

### `PreEncoded<M>`

An `Encodable<M>` body that wraps already-encoded protobuf bytes, with a `PhantomData<fn() -> M>` witness so the type system enforces the bytes belong to the right RPC. `PreEncoded::from_view(&view)` constrains `V: MessageView<'a, Owned = M>` for a compile-time match.

**Proto-only by design.** JSON clients receive `unimplemented`, same as the existing per-output-type view-body impls. ViewEncode is a high-throughput optimization; JSON is not a high-throughput codec, and supporting both would add a runtime format check on every encode for no realistic consumer.

```rust
async fn resolve(&self, _ctx: RequestContext, req: ...)
    -> ServiceResult<PreEncoded<ResolveResponse>>
{
    let snapshot = self.store.load();
    let view = build_view_from_snapshot(&snapshot, &req);
    Response::ok(PreEncoded::from_view(&view))
}
```

### `type Item: Encodable<Res>` on streaming handlers

`StreamingHandler`, `BidiStreamingHandler`, and their `View*` variants gain `type Item: Encodable<Res> + Send + 'static` and return `ServiceStream<Self::Item>` — bringing them to parity with unary `Handler::Body` (which got the same treatment in 0.4.0). `encode_body_stream` / `encode_response_stream` are generalized over `B: Encodable<Res>`. Codegen emits the streaming trait method as `ServiceStream<impl Encodable<Out> + Send + use<Self>>`.

Vestigial `Res: Serialize` bounds dropped from the streaming wrappers — they only call `Encodable::encode`, which is codec-agnostic.

## Breaking — but the practical surface is narrow

This is breaking under semver, but the practical blast radius is much smaller than the trait change suggests:

- **`impl <GeneratedServiceTrait>` blocks compile unchanged.** The generated trait declares `ServiceStream<impl Encodable<Out> + Send + use<Self>>` (RPITIT). Existing handler impls returning `ServiceStream<Out>` still compile via RPITIT refinement — the same mechanism the unary `Body` path has used since 0.4.0, with the same `allow(refining_impl_trait)` already documented in 0.4.1. Verified: `tests/streaming/src/lib.rs`'s pre-existing handlers were not touched in this diff.
- **`*_handler_fn` callers compile unchanged.** The helpers infer `type Item` from the closure return type.
- **Only hand-rolled `impl StreamingHandler` / `impl BidiStreamingHandler` blocks need an edit** — a one-line `type Item = Res;`. These are the low-level escape-hatch traits behind `Router`, not the primary handler surface. At the time of this PR there are two such impls across all known consumers.
- **Checked-in `protoc-gen-connect-rust` output must be regenerated** — same regeneration footgun documented in 0.4.0; `connectrpc-build` users are unaffected.

A non-breaking alternative was considered (a separate `Router::route_server_stream_pre_encoded(...)` registration path that bypasses `StreamingHandler` entirely). It works, but permanently doubles the registration API surface and needs a per-method codegen flag, all to avoid a 2-line migration. The `type Item` shape is the cleaner long-term API and the precedent (unary `Body` in 0.4.0) was a bigger break and shipped without issue.

## Size

~700 lines net (excluding generated). Over the usual 250 guideline because the change is coupled: the trait, `encode_body_stream`, the dispatcher, the codegen, and the regenerated checked-in code all have to move together — splitting would leave intermediate states uncompilable.

## Tests

- Unit: `PreEncoded::proto`/`from_view` round-trip + JSON-unimplemented + typed-witness assertion.
- Streaming: `streaming_handler_item_is_inferred_from_closure` (compile-time inference check), `encode_body_stream` with `PreEncoded` items.
- Integration: `server_stream_pre_encoded_round_trip` in `tests/streaming/` — registers a streaming handler with `type Item = PreEncoded<EchoResponse>`, runs through the full server → axum → client → decode chain.
- All workspace tests pass; clippy `-D warnings` clean; nightly fmt clean.

## Notes for reviewers

- **Hand-rolled `streaming_handler_fn` users** must turbofish `Res` (`streaming_handler_fn::<_, _, Req, Res, PreEncoded<Res>>(...)`) since the generic body type makes it non-inferable. The codegen-driven path (the common case) hides this. Documented on the helper's doc comment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,16 +20,21 @@ increment the patch version.
   view with a non-`'static` lifetime can't cross the handler boundary;
   `PreEncoded` carries the bytes across instead.
 
-  The `M` type parameter is a compile-time witness:
-  `PreEncoded::from_view(&view)` enforces `M = MView::Owned` so the
-  bytes are guaranteed to decode as `M`; `PreEncoded::proto(bytes)`
-  wraps already-encoded bytes you've produced yourself and takes `M`
-  on trust.
+  The `M` type parameter is a compile-time witness. Three construction
+  paths, in decreasing order of guarantee:
+  `PreEncoded::from_message(&m)` (also `From<&M>`/`.into()`) encodes an
+  owned `M` and the receiver type is the witness;
+  `PreEncoded::from_view(&view)` enforces `M = MView::Owned`;
+  `PreEncoded::from_bytes_unchecked(bytes)` wraps already-encoded bytes
+  from elsewhere — a cache, a blob store, a sidecar — and takes `M` on
+  trust (in debug builds it also decodes once as a `debug_assert!`).
 
-  Proto-only, like the per-output-type view-body impls. JSON clients
-  receive an `unimplemented` error; if your service must support JSON,
-  build the owned message and return it (or `MaybeBorrowed::Owned`) on
-  every path.
+  Optimized for the proto codec, where the bytes pass through verbatim.
+  JSON requests fall back to decoding the proto bytes as `M` and
+  re-serializing — slow, but a working response rather than a runtime
+  `unimplemented` error. Services with significant JSON traffic should
+  build and return the owned message (or `MaybeBorrowed::Owned`) so the
+  codec layer can skip the proto round-trip.
 
 ### Breaking
 
@@ -53,8 +58,9 @@ need a one-line edit.
   service trait or the `*_handler_fn` closure helpers, neither of which
   is affected — codegen handles the new shape and the helpers infer
   `type Item` from the closure return. Hand-rolled `impl
-  StreamingHandler` blocks must add `type Item = Res;`. (At the time of
-  this change there are two such impls across all known consumers.)
+  StreamingHandler` blocks must add `type Item = Res;`. We surveyed the
+  in-tree consumers and found two; if you have hand-rolled impls,
+  expect a single one-line addition per impl block.
 
 - **Generated server-streaming and bidi-streaming trait methods now
   declare `ServiceStream<impl Encodable<Out> + Send + use<Self>>`**
@@ -71,15 +77,14 @@ need a one-line edit.
   regenerate** (the same regeneration footgun documented in 0.4.0).
   `connectrpc-build` users (build.rs) are unaffected.
 
-- **`encode_response_stream` / `encode_body_stream` gain a `B` type
-  parameter** for the stream item type. `Res` is no longer derivable
-  from the item type alone (the closure return type doesn't constrain
-  which `Res` to project through `Encodable<Res>`), so callers must
-  turbofish:
-  `encode_response_stream::<Res, _, _>(s, format)`. The generated
-  dispatcher and route-registration code does this; only consumers that
-  call `dispatcher::codegen::encode_response_stream` directly need to
-  update. We are not aware of any.
+- **`dispatcher::codegen::encode_response_stream` gains a `B` type
+  parameter** for the stream item type. The generated dispatcher and
+  route-registration code passes `Res` explicitly because the trait
+  method's stream item is the *opaque* `impl Encodable<Out>` (RPITIT),
+  which can't be unified against the `Encodable<Res>` impls. Only
+  consumers that call `dispatcher::codegen::encode_response_stream`
+  directly need to turbofish `encode_response_stream::<Res, _, _>(s,
+  format)`. We are not aware of any.
 
 ## [0.4.2] - 2026-05-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,29 +33,43 @@ increment the patch version.
 
 ### Breaking
 
+These are breaking under semver but the practical blast radius is
+narrow. The common consumer surfaces — `impl <GeneratedServiceTrait>`
+blocks and `*_handler_fn` registrations — compile unchanged. Only
+hand-rolled `impl StreamingHandler`/`impl BidiStreamingHandler` blocks
+and direct callers of `dispatcher::codegen::encode_response_stream`
+need a one-line edit.
+
 - **Streaming handler traits gain `type Item: Encodable<Res>`** and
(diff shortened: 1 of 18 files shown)


iainmcgin:
[claude code] Reviewed in detail with @iainmcgin — the abstraction is sound and the breaking-change blast radius assessment in the description is accurate. We pushed three commits on top (`b6fc166`..`d5dc61f`) addressing the review findings rather than relitigating in comments. Semantic diff below; happy to discuss or revert any of it.

## `PreEncoded` constructor surface — `b6fc166`

**`PreEncoded::proto` → `PreEncoded::from_bytes_unchecked`.** The constructor wraps unvalidated bytes; the new name carries that contract at the call site (`str::from_utf8_unchecked` precedent). Added a `debug_assert!` that decodes the bytes once in debug builds to surface mismatches early.

**Added `PreEncoded::from_message(&m)` and `From<&M>`.** This was a gap: a stream with `type Item = PreEncoded<M>` mixing a cache-hit path (owned `M`) with a cache-miss path (view from a snapshot) had no checked way to produce the owned-message item — `from_bytes_unchecked(m.encode_to_bytes())` was the only option, which throws away a guarantee for no reason. The triple now reads as a deliberate spectrum:

```rust
PreEncoded::from_message(&m)            // receiver type is the witness
PreEncoded::from_view(&view)            // MessageView::Owned = M is the witness
PreEncoded::from_bytes_unchecked(bytes) // no witness — bytes from a cache/store/sidecar
```

**JSON codec fallback.** This is the one design departure from your "Proto-only by design" stance, so flagging it explicitly. The `Encodable<M> for PreEncoded<M>` impl now decodes the proto bytes back to `M` and re-serializes for `CodecFormat::Json` instead of returning `unimplemented`. Rationale: the codec is selected per-request by the client's `Content-Type`, so a `PreEncoded` handler on a JSON-capable router *will* receive JSON requests, and "works slowly" beats "fails at runtime in a way nothing catches at registration." Two real consequences are documented in the new `# Codec behaviour` section:

- *Codec-dependent fidelity.* The proto path is byte-exact; the JSON path round-trips through `M`. Unknown fields and non-canonical encodings survive proto and not JSON. Only relevant for `from_bytes_unchecked` bytes from external sources — `from_message`/`from_view` can't carry unknown fields by construction. Pinned with a test.
- *Cost is selected by the client.* A client sending `application/json` shifts that request onto the slow decode-reserialize path (per item for streaming). The doc points services that want to enforce proto-only at the middleware layer (`415 Unsupported Media Type` before the body is read) rather than relying on the body type to reject it.

The `Encodable` impl bound tightens to `M: Message + Serialize`, matching the blanket impl for owned messages — not observably narrower for any buffa-generated message.

If you'd rather keep `unimplemented` as the default and make the fallback opt-in, that's a small change — say the word.

**Coherence note correction.** The original comment claimed the `impl<M> Encodable<M> for PreEncoded<M>` block depends on `PreEncoded` never implementing `Message`/`Serialize`. That's not actually the constraint — the two impls are non-overlapping for structural reasons (`T = PreEncoded<T>` is infinite) regardless of what `PreEncoded` implements. We verified this empirically and rewrote the comment with the correct argument. Worth a look since it's load-bearing reasoning for a future reader.

**`encode_body_stream` / `encode_response_stream` dedup.** They were byte-identical after the `B`-generic generalization; `encode_body_stream` is now a thin forward.

**`streaming_handler_fn` doc precision.** The original doc says a turbofish is required for `PreEncoded` items. We tested this and it's narrower than that: inference works whenever the closure pins `B` concretely (`from_view`, `from_message`, or `PreEncoded::<M>::from_bytes_unchecked(...)`). It only fails for `PreEncoded::from_bytes_unchecked(bytes)` with no `::<M>`, and the simpler fix there is to name `M` at the construction site, not to spell out a 5-element turbofish on the helper. The codegen path is genuinely different (the trait method's item is the *opaque* RPITIT, which can't unify), and the codegen turbofishes correctly. The `streaming_handler_item_is_inferred_from_closure` test now demonstrates the no-turbofish `PreEncoded` case.

## Codegen trait docstring — `19f7965`

The generated `# Implementing handlers` section listed `Out` / view types / `MaybeBorrowed` body alternatives but not `PreEncoded` — the type a user needs precisely when they hit a `'static` lifetime error trying to return a borrowing view. Added it, plus a paragraph explaining why streaming methods say `+ use<Self>` (vs `+ use<'a, Self>` on unary): stream items must be `'static`, `PreEncoded` is the escape hatch. Regenerated the four checked-in consumers; only the docstring changed.

## Bidi integration test — `d5dc61f`

`PreEncodedEchoService::bidi_stream` now yields `PreEncoded` items (same build-view-encode-yield-bytes shape as `server_stream`), and `bidi_stream_pre_encoded_round_trip` covers the `BidiStreamingViewHandlerWrapper` chain end-to-end.

## Verification

`cargo test --workspace` 387 passed, `cargo clippy --workspace --all-features --all-targets -- -D warnings` clean, nightly `cargo fmt --check` clean, `task generate:all` produces no diff in `*/generated/connect/`. The pre-existing `EchoServiceImpl` impls in `tests/streaming/src/lib.rs` returning `ServiceStream<EchoResponse>` compile unchanged — confirms the RPITIT refinement claim from the PR description.

iainmcgin (approved):

Issue #91, opened 2026-05-06
Router::route_* manual routing is stringly typed and surfaces in rustdoc before the codegen path


iainmcgin:
`Router` exposes a family of stringly-typed `route_*` methods:

```rust
pub fn route(self, service_name: &str, method_name: &str, handler: ...) -> Self
pub fn route_view_idempotent(self, service_name: &str, method_name: &str, handler: ...) -> Self
pub fn route_view_server_stream(...)
pub fn route_view_client_stream(...)
pub fn route_view_bidi_stream(...)
// ... 8 variants total
```

A typo in `service_name` or `method_name` produces a runtime `Unimplemented` for clients — not a compile error. The generated `<Service>Ext::register` path avoids this entirely (it passes the `*_SERVICE_NAME` constants), but the manual `route_*` family is what shows up first in `Router`'s rustdoc, so it's the API new users see and the one that invites the bug.

These methods are essentially codegen plumbing. They need to remain `pub` so generated code can call them, but they shouldn't be the documented entry point for end users.

## Proposed direction

- `#[doc(hidden)]` on the `route_*` family, with a doc note pointing at the generated `register` extension trait as the supported path.
- If a manual-routing escape hatch is genuinely needed (e.g. proxies that route without codegen), expose one that takes typed arguments — a `MethodDescriptor`/`Spec` struct rather than two `&str`s — so a typo is at least a `MethodDescriptor` construction error rather than a silent `Unimplemented`.
- Consider collapsing the 8 `route_view_*` variants into one `route_view::<...>(spec, handler)` form that takes an `RpcKind` enum, reducing the surface even on the codegen side.

## Scope

Small (`#[doc(hidden]` + doc updates), or medium if we also collapse the variants.

iainmcgin:
[claude code] Resolved by #162 (low-level `route_*` helpers hidden from rustdoc, docs point at generated `<Service>Ext::register`) plus #164 (typed `add_service` registration), both merged 2026-06-20. Auto-close didn't fire because the PRs landed through the merge queue.

Issue #89, opened 2026-05-06
RequestContext field exposure: decide #[non_exhaustive] / accessor policy before adding fields


iainmcgin:
`RequestContext` is a `pub`-fields struct without `#[non_exhaustive]`:

```rust
pub struct RequestContext {
    pub headers: HeaderMap,
    pub deadline: Option<Instant>,
    pub extensions: http::Extensions,
}
```

This means adding a field is a SemVer-major break for anyone who constructed a `RequestContext` literal (e.g. in tests). We have already grown this struct once (the mTLS work added connection-scoped data via `extensions` precisely to avoid a new field), and the ergonomics review identified several more candidate fields: service name, method name, RPC kind, negotiated protocol — all of which generic middleware/logging needs and which neither tonic's `Request<T>` nor today's `RequestContext` carry.

We should decide the evolution policy now, before 1.0:

**Option A — `#[non_exhaustive]` + accessors.** Make fields `pub(crate)`, add `headers()`, `headers_mut()`, `deadline()`, `extensions()`, `extensions_mut()` accessors plus a builder for tests. Future fields are additive. This is the conventional Rust-library answer.

**Option B — commit to the open struct.** Keep fields `pub` and accept that any new request-scoped data must flow through `extensions`. Cheaper today, but pushes useful fields (method name, protocol) into a typed-map lookup with no discoverability.

Either choice should also add accessor helpers for the well-known extension types so users don't write `.extensions.get::<PeerAddr>().unwrap()` (which compiles, runs in tests, and panics in production because the `Server` inserts `PeerAddr` but axum integrations may not):

```rust
impl RequestContext {
    pub fn peer_addr(&self) -> Option<SocketAddr>;
    pub fn peer_certs(&self) -> Option<&[CertificateDer]>;
}
```

## Scope

Small but SemVer-relevant. Best landed before any field is added under pressure.

iainmcgin:
[claude code] Resolved by #101 — `RequestContext` is now `#[non_exhaustive]` with accessor methods (`path()`, `spec()`, `protocol()`, etc.), settling the field-exposure policy.

PR #103, opened 2026-05-14
service: DeadlinePolicy for server-side timeout moderation

### LINKED ISSUE #92: Deadline is parsed but not enforced — handler runs to completion after Connect-Timeout-Ms lapses
The runtime parses `Connect-Timeout-Ms` / `grpc-timeout` headers into `RequestContext::deadline: Option<Instant>`, but does not cancel the handler future when the deadline lapses. The handler must check `ctx.deadline` itself.

This differs from tonic and connect-go, both of which cancel the handler when the deadline passes (tonic drops the future; connect-go's `context.Context` is cancelled). A user who writes a slow handler and relies on the protocol timeout for protection will find their handler running to completion, holding DB connections, locks, and worker-pool slots, only for the response to be discarded because the client has already gone away. Under load this turns a slow downstream into a cascading resource exhaustion.

It's a defensible design choice — a handler may legitimately want to commit a transaction it has started even if the client gave up — but it's a surprising default and should at minimum be a documented, explicit choice.

## Proposed direction

- Default to enforcing the deadline: wrap the handler future in `tokio::time::timeout(deadline - now, handler)` and translate the timeout into a `DeadlineExceeded` error. This matches the behavior of every other framework in the comparison set and the gRPC spec's intent.
- Add an opt-out: `Server::enforce_deadlines(false)` (and an equivalent for the axum/tower integration path, perhaps a `RouterConfig` knob) for services that need to outlive the caller.
- Document the behavior and the opt-out in the guide's "Production hardening" section.
- Consider exposing the remaining time (`ctx.time_remaining() -> Option<Duration>`) so handlers can budget downstream calls, similar to gRPC-Go's `ctx.Deadline()` idiom.

## Open questions

- For client/bidi streaming handlers, should the deadline apply to the whole stream lifetime or be resettable per message?
- Whether cancellation should be "drop the future" (Rust idiom, may leak partial work) or "signal and grace period" (closer to Go).

## Scope

Medium — touches the dispatcher hot path, needs careful testing against the conformance suite's timeout cases.

iainmcgin:
Fixes #92.

Adds `DeadlinePolicy` for server-side moderation of client-asserted RPC deadlines. A client controls the per-request timeout via `Connect-Timeout-Ms` / `grpc-timeout`; without a policy the server trusts that value verbatim — a 1ms request cancels the handler mid-write, a 24h request holds a worker indefinitely, no header means no bound, and once a streaming handler returns its stream the items flow unbounded regardless of the header.

`DeadlinePolicy` clamps the client value to a server-controlled `[min, max]`, supplies a default when the client asserts nothing (or sends garbage), and optionally extends enforcement to streaming response bodies and to the gap between yielded items.

```rust
let policy = DeadlinePolicy::new()
    .with_min(Duration::from_millis(5))
    .with_max(Duration::from_secs(30))
    .with_default(Duration::from_secs(10))
    .with_enforce_on_streams(true);
ConnectRpcService::new(router).with_deadline_policy(policy)
// or: Server::new(router).deadline_policy(policy)
```

`DeadlinePolicy::new()` is a strict no-op — no clamping, no default, no stream enforcement. **Existing services that don't set a policy see exactly the previous behavior**, so this is purely additive and safe in 0.5.0. The recommended production starting point is at least `with_max()` and `with_default()`.

`min` is `Duration` (not `Option`) because zero is a natural identity for a floor; `max`/`default` are `Option<Duration>` because there's no clean identity for them. The asymmetry is deliberate.

Stream-body enforcement (when `enforce_on_streams` is set) is a pin-projected wrapper applied to server- and bidi-streaming response bodies. Unary and client-streaming responses are already bounded by the existing `tokio::time::timeout` on the handler future. `inter_message_timeout` re-arms a per-item timer to detect stalled streams independent of the absolute deadline.

When a clamp changes the client value, a `tracing::debug!` event fires on target `connectrpc::deadline` with the path and before/after durations.

### Notes for reviewers

- `Server::with_deadline_policy()` is a new delegating builder (renamed from `deadline_policy()` after review to match `ConnectRpcService::with_deadline_policy` and avoid colliding with the same-named getter on `ConnectRpcService`). `Server` does not yet proxy `ConnectRpcService::with_compression()` / `with_compression_policy()` / `with_limits()` — that's a pre-existing gap; the `Server::from_service(ConnectRpcService::new(...).with_*(...))` path is the canonical way to set those today. Worth a follow-up to either add the missing proxies or document `from_service` as the configuration path.
- `enforce_on_streams` and `inter_message_timeout` are independent knobs — either alone wraps the stream. Both default off.
- `RequestContext.deadline` reflects the *moderated* value once a policy is set; the field doc cross-links `DeadlinePolicy`. PR #101 (issue #89) replaces the field with a `deadline()` accessor and adds `time_remaining()` — there will be a small merge conflict in `response.rs` between this PR and #101, expected and trivial.
- I have not run the conformance suite locally (the runner wasn't downloaded in the dev environment). Since `DeadlinePolicy::new()` is a strict identity (`moderate` returns the input unchanged; `enforce_on_response_stream` returns the stream unchanged), conformance behavior is provably unaffected without a configured policy. CI runs the full suite.
- Per-route deadline policy is deferred to a follow-up coupled to the typed routing surface (#91). A per-path string map now would be a second stringly-typed route surface that #91 wants to remove.
- Net change in `connectrpc/src/`: ~250 lines excluding tests and doc comments. The `deadline.rs` file is large because of doc-comment density on every public knob.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,38 @@ increment the patch version.
   build and return the owned message (or `MaybeBorrowed::Owned`) so the
   codec layer can skip the proto round-trip.
 
+- **`DeadlinePolicy`** — server-side moderation of client-asserted RPC
+  deadlines. Clients control the per-request timeout via
+  `Connect-Timeout-Ms` / `grpc-timeout`; without a policy the server
+  trusts that value verbatim. `DeadlinePolicy` clamps the asserted
+  timeout to a server-controlled `[min, max]` range, applies a default
+  when the client asserts nothing (or sends an unparseable header), and
+  optionally extends enforcement to streaming bodies.
+
+  Configure via `ConnectRpcService::with_deadline_policy(...)` (axum /
+  tower) or `Server::with_deadline_policy(...)`. `DeadlinePolicy::new()`
+  is a no-op policy that preserves the prior default behavior — no
+  clamping, no default, streaming bodies unbounded once the handler
+  returns. Existing services see no change without an explicit opt-in.
+  Recommended production starting point: set at least `with_max(...)` to
+  bound worker lifetime and `with_default_timeout(...)` to your SLA.
+
+  Two independent opt-in extensions, both off by default:
+  - `with_enforce_on_streams(true)` wraps server- and bidi-streaming
+    response bodies so the next item after the deadline is a
+    `deadline_exceeded` error and the stream ends. Unary and
+    client-streaming responses were already bounded by the handler
+    future timeout; this closes the streaming-body gap.
+  - `with_inter_message_timeout(d)` cuts off a stream that goes longer
+    than `d` between yielded items (a stalled handler waiting on a slow
+    upstream). Takes effect whenever set, with or without
+    `with_enforce_on_streams`.
+
+  When clamping changes a client value, a `tracing::debug!` event fires
+  on target `connectrpc::deadline` with the path and before/after
+  durations. Per-route deadline policy is a planned follow-up coupled
+  to the typed routing surface (#91).
+
 ### Breaking
 
 These are breaking under semver but the practical blast radius is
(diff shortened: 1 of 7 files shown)


rpb-ant (approved):
lgtm :shipit:

Issue #110, opened 2026-05-18
Expose procedure constants to the generated code


DevExzh:
For the code generated by the tool `protoc-gen-connect-go`, it exposes all the procedure strings. For instance, the `BookService` below:

```protobuf
service BookService {
  rpc CreateBook(CreateBookRequest) returns (CreateBookResponse);
  rpc GetBook(GetBookRequest) returns (GetBookResponse);
  rpc ListBooks(ListBooksRequest) returns (ListBooksResponse);
  rpc UpdateBook(UpdateBookRequest) returns (UpdateBookResponse);
  rpc DeleteBook(DeleteBookRequest) returns (DeleteBookResponse);
  rpc BorrowBook(BorrowBookRequest) returns (BorrowBookResponse);
  rpc ReturnBook(ReturnBookRequest) returns (ReturnBookResponse);
}
```

would generate these Go consts:

```go
const (
	// BookServiceName is the fully-qualified name of the BookService service.
	BookServiceName = "example.library.v1.BookService"
)

// These constants are the fully-qualified names of the RPCs defined in this package. They're
// exposed at runtime as Spec.Procedure and as the final two segments of the HTTP route.
//
// Note that these are different from the fully-qualified method names used by
// google.golang.org/protobuf/reflect/protoreflect. To convert from these constants to
// reflection-formatted method names, remove the leading slash and convert the remaining slash to a
// period.
const (
	// BookServiceCreateBookProcedure is the fully-qualified name of the BookService's CreateBook RPC.
	BookServiceCreateBookProcedure = "/example.library.v1.BookService/CreateBook"
	// BookServiceGetBookProcedure is the fully-qualified name of the BookService's GetBook RPC.
	BookServiceGetBookProcedure = "/example.library.v1.BookService/GetBook"
	// BookServiceListBooksProcedure is the fully-qualified name of the BookService's ListBooks RPC.
	BookServiceListBooksProcedure = "/example.library.v1.BookService/ListBooks"
	// BookServiceUpdateBookProcedure is the fully-qualified name of the BookService's UpdateBook RPC.
	BookServiceUpdateBookProcedure = "/example.library.v1.BookService/UpdateBook"
	// BookServiceDeleteBookProcedure is the fully-qualified name of the BookService's DeleteBook RPC.
	BookServiceDeleteBookProcedure = "/example.library.v1.BookService/DeleteBook"
[...]

Issue #87, opened 2026-05-06
Observability: per-RPC tracing spans and OpenTelemetry integration


iainmcgin:
There is currently no per-RPC observability hook. The runtime emits `tracing` events at `debug`/`trace` level for connection state and envelope decoding, but no span scoped to an individual RPC. Users who want a span carrying `rpc.system`, `rpc.service`, `rpc.method`, and `rpc.grpc.status_code` (the OpenTelemetry RPC semantic conventions) have nowhere to put it.

The recommended middleware pattern (`tower_http::trace::TraceLayer::new_for_http()` from the middleware example) produces HTTP spans (`POST /pkg.Service/Method 200`), not RPC spans. The HTTP layer doesn't know the RPC kind, can't see the Connect/gRPC error code (which is in the trailer or JSON body, not the HTTP status for most codes), and doesn't compose with downstream services that propagate W3C trace context through metadata.

For comparison: connect-go ships `connectrpc.com/otelconnect` (an `Interceptor` that creates spans and metrics per the OTel RPC conventions and propagates trace context); tonic users reach for `tonic-tracing-opentelemetry`'s `OtelGrpcLayer`. Both depend on having a typed RPC-level interception point.

## Proposed direction

Once a typed RPC-level interceptor surface exists (tracked separately), ship a `connectrpc-otel` companion crate (or a feature-gated module) providing:

- a server-side interceptor that opens a span per RPC with `rpc.system = "connect_rpc"` (or `"grpc"`/`"grpc_web"` depending on the negotiated protocol), `rpc.service`, `rpc.method`, and records `rpc.grpc.status_code` on completion
- a client-side interceptor that propagates W3C `traceparent`/`tracestate` (and `grpc-trace-bin` for gRPC peers) via request headers and opens a client span
- optional metrics (request count, duration histogram, message size histogram) keyed by service/method/code

Even before the interceptor work lands, two cheap improvements are possible:

- expose service name, method name, and negotiated protocol on `RequestContext` (or a `Spec` struct) so a user-written tower layer can at least get the labels right by reading extensions
- a `TracingLayer` that wraps the existing HTTP-level path in a span named `pkg.Service/Method` rather than `POST /pkg.Service/Method`

## Scope

- depends on typed interceptor surface (separate issue)
- `connectrpc-otel` crate or `otel` feature, server + client interceptors
- example showing OTLP export end to end

PR #112, opened 2026-05-18
spec: add Spec, StreamType, IdempotencyLevel; thread through dispatch

aknott-ant:
Add static per-method metadata that handlers (and a future interceptor surface) can read without re-parsing the request URL.

- New `connectrpc::spec` module: `Spec` (procedure path, stream type, idempotency level, client/server side), `StreamType`, `IdempotencyLevel`, `SpecOrigin`. All `Copy` and `'static`, with `const fn` constructors (`Spec::server(...)` / `Spec::client(...)`) so codegen can emit `Spec` constants in `.rodata`.
- `MethodDescriptor` gains `spec: Option<Spec>`, becomes `#[non_exhaustive]`, and grows `from_kind`/`with_idempotent`/`with_spec` const builders. The dynamic `Router` returns `spec: None` (its keys are owned `String`s, not `&'static str`).
- `RequestContext` gains `spec: Option<Spec>` and `protocol: Option<Protocol>` with `with_spec`/`with_protocol` builders.
- The dispatch path in `service.rs` threads the looked-up `Spec` and the negotiated `Protocol` into every `RequestContext` (Connect unary, gRPC unary fast path, server/client/bidi streaming).
- `connectrpc-codegen` emits a `pub const FOO_SERVICE_BAR_SPEC: Spec` per method and references it from the generated `Dispatcher::lookup`.
- CHANGELOG `[Unreleased]` entries and a `docs/guide.md` "Static method metadata (`Spec`)" subsection covering `ctx.spec()`, the `Router`-returns-`None` gotcha, and the generated `<SERVICE>_<METHOD>_SPEC` constants.

`Protocol` lives on `RequestContext` rather than `Spec` because it is connection-scoped, not method-scoped (matching connect-go's `Spec`/`Peer` split); a `const Spec` cannot carry a runtime-negotiated field.

`SpecOrigin` is an enum (`Server` / `Client`) rather than connect-go's `IsClient bool` — closed two-valued domains read better as enums in Rust (`spec.origin == SpecOrigin::Client`), the constructors `Spec::server(...)` / `Spec::client(...)` carry the side at the call site without a builder, and the `Spec` prefix disambiguates from a request "origin" in the HTTP/CORS sense.

Conformance: 3600/3600 (no behavioral change to wire handling).

Refs #87, #110.


iainmcgin:
recheck

iainmcgin:
recheck

aknott-ant:

iainmcgin (approved):

PR #113, opened 2026-05-18
payload: AnyMessage and Payload — type-erased, lazily-decoded message bodies

aknott-ant:
> **Stacked on #112** (`aknott/interceptors-pr1-spec`).
>
> This PR targets `main`, so its diff currently also includes the #112
> commit. **Review only the top commit** — _"payload: AnyMessage and
> Payload — type-erased, lazily-decoded message bodies"_. Once #112 merges,
> this branch will be rebased and the diff will reduce to just that commit.
> Kept as a draft until #112 lands.

## Summary

Foundation for RPC interceptors (PR 2 of the interceptor series). Most
interceptors only read `Spec` and headers, never the message body, so
eagerly decoding for every call would tax the common case. This PR adds the
type-erased, lazily-decoded message machinery the interceptor surface builds
on.

- **`AnyMessage`**: object-safe surface for type-erased messages, with a
  blanket impl over every `T: Message + Serialize`. No codegen required;
  covers every owned message `buffa` generates.
- **`Payload`**: wire `Bytes` (refcounted, always available) + `CodecFormat`
  + a `OnceLock<Box<dyn AnyMessage>>` lazy decode cache + an optional
  replacement set by `set_message`. Typed access via `message::<M>()`
  (decode-once, proto and JSON) or `view::<V>()` (zero-copy proto view;
  errors for JSON wires, which cannot back a zero-copy view). `encoded()`
  returns the original bytes or re-encodes the replacement.

## Notes

`Payload` is intentionally not `Clone` — a clone would either drop or
duplicate the decode cache. Errors from the typed accessors that indicate
caller-side type bugs use `Internal` (a server-side programming error); the
JSON-wire `view()` error uses `FailedPrecondition` since it is a documented
limitation, not a bug.

## Interceptor series

1. #112 — `Spec` / `StreamType` / `IdempotencyLevel`
2. **(this PR)** — `AnyMessage` / `Payload`
3. (next) — `Interceptor` trait, `Next`, server registration

## Verification

Run on the rebased branch (buffa 0.6.0):

- `task generate:all` — no generated-code drift
- `task lint` — pass
- `task test` — pass, 0 failed

### DIFF
--- a/connectrpc/src/client/mod.rs
+++ b/connectrpc/src/client/mod.rs
@@ -314,6 +314,16 @@ enum HttpClientInner {
 #[cfg(feature = "client")]
 #[cfg_attr(docsrs, doc(cfg(feature = "client")))]
 impl HttpClient {
+    /// Returns a builder for configuring connector-level options before
+    /// choosing a transport flavour.
+    ///
+    /// `HttpClient::plaintext()` is equivalent to
+    /// `HttpClient::builder().plaintext()`, and likewise for the other
+    /// constructors.
+    pub fn builder() -> HttpClientBuilder {
+        HttpClientBuilder::default()
+    }
+
     /// Create a **plaintext** HTTP client. Only for `http://` URIs.
     ///
     /// Errors at send-time if given an `https://` URI — use
@@ -323,17 +333,7 @@ impl HttpClient {
     /// over cleartext. TCP_NODELAY is enabled to avoid Nagle + delayed ACK
     /// latency on small messages.
     pub fn plaintext() -> Self {
-        use hyper_util::client::legacy::Client;
-        use hyper_util::client::legacy::connect::HttpConnector;
-        use hyper_util::rt::TokioExecutor;
-
-        let mut connector = HttpConnector::new();
-        connector.set_nodelay(true);
-        let client = Client::builder(TokioExecutor::new()).build(connector);
-
-        Self {
-            inner: HttpClientInner::Plain(client),
-        }
+        Self::builder().plaintext()
     }
 
     /// Create a **plaintext** HTTP client with HTTP/2 prior-knowledge (h2c) only.
@@ -350,19 +350,7 @@ impl HttpClient {
     /// it has honest `poll_ready` and composes with `tower::balance`. This
     /// method pins you to one connection per host with no way to scale out.
     pub fn plaintext_http2_only() -> Self {
-        use hyper_util::client::legacy::Client;
-        use hyper_util::client::legacy::connect::HttpConnector;
-        use hyper_util::rt::TokioExecutor;
-
-        let mut connector = HttpConnector::new();
-        connector.set_nodelay(true);
-        let client = Client::builder(TokioExecutor::new())
-            .http2_only(true)
-            .build(connector);
-
-        Self {
-            inner: HttpClientInner::Plain(client),
-        }
+        Self::builder().plaintext_http2_only()
     }
 
     /// Create a **TLS** HTTP client. Only for `https://` URIs.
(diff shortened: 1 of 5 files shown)


aknott-ant:

iainmcgin:
[claude code] Reviewed `payload.rs` after rebasing onto #116 (clean rebase, 317 tests pass, no conflicts — pushed to your branch with `--force-with-lease` so the diff drops the duplicated #112 commit).

The lazy-decode design is well-built — the `OnceLock` probe-then-set is reasoned through correctly, the `!Clone` rationale is sound, the doc comments are excellent. Nothing architecturally blocking. Two things I'd address before merge, four worth a look.

## High

### H1 — `view()` should return `Internal`, not `FailedPrecondition`, for JSON wires

`payload.rs:223–227`. The error escapes to the wire if the interceptor doesn't catch it, and from the *peer's* perspective they sent a perfectly valid JSON request and got `FAILED_PRECONDITION` with a message about server internals. gRPC's code semantics say `FAILED_PRECONDITION` is "the system is in a wrong state — retry won't help until *you* change something" — but the only thing the client could change is the codec, and Connect gives them no way to discover that's what's wrong.

The actual fault is server-side: an interceptor called `view()` without checking `format()` first. That's a programming error → `Internal`. The commit-message rationale ("a documented limitation, not a bug") describes the *SDK API contract*, but the error code is part of the *server-client contract*, and those two audiences see it differently.

Proposed:

```rust
if self.format != CodecFormat::Proto {
    return Err(ConnectError::internal(
        "Payload::view requires a proto-encoded wire; use Payload::message for JSON",
    ));
}
```

with a doc note that this error escapes to the wire as a 500, so callers should branch on `format()` rather than treat the error as a fallback signal:

```rust
/// # Errors
///
/// Returns [`internal`](ConnectError::internal) for JSON-encoded wires:
/// JSON cannot back a zero-copy proto view. This is a server-side
/// programming error and escapes to the peer as a 500 if uncaught —
/// branch on [`format()`](Payload::format) and call
/// [`message()`](Payload::message) for JSON wires instead. Also returns
/// [`invalid_argument`](ConnectError::invalid_argument) if the wire bytes
/// fail to decode as `V` (peer-supplied bytes), and `internal` if a
/// replacement set with [`set_message`](Payload::set_message) fails to
/// re-encode (server-supplied bytes).
```

(That last sentence also documents the existing — and correct — asymmetry where wire-bytes decode failure is `invalid_argument` and replacement decode failure is `internal`. Worth pinning in the doc so it doesn't read like an oversight.)

### H2 — `Send + Sync` bound discipline is inconsistent

`payload.rs:69–72` writes the blanket impl bound as `T: Message + Serialize + Send + Sync + 'static`. `payload.rs:158–160` writes `message<M>`'s bound as `M: Message + Serialize + DeserializeOwned + 'static` — no `Send + Sync` — and relies on `buffa::Message: Send + Sync` (supertraits) for `Box::new(m)` to coerce to `Box<dyn AnyMessage>` at line 182. Both compile, but they encode opposite assumptions about what `Message` carries.

Pick one. I'd drop `+ Send + Sync` from the blanket impl since `buffa::Message` already requires both as supertraits, which keeps the bounds short everywhere:

```rust
impl<T> AnyMessage for T
where
    T: Message + Serialize + 'static,
```

The other direction — adding `+ Send + Sync` to `message<M>`'s `where` clause — is fine too and gives a clearer diagnostic at the call site if buffa ever drops the supertrait. The thing I'd avoid is having both styles in the same file 90 lines apart.

## Medium

### M1 — `bytes()` next to `encoded()` is a naming footgun

After `set_message`, `bytes()` returns the *original* (pre-replacement) bytes and `encoded()` returns what's about to be sent. "Bytes" and "encoded" are near-synonyms in protobuf vocabulary, and an author reaching for "the bytes going on the wire" will not reliably pick `encoded()`. The doc on `bytes()` says "regardless of any replacement," which is technically true but easy to skim past.

At minimum, make the warning unmissable:

```rust
/// The original wire bytes the peer sent, **ignoring** any replacement
/// set with [`set_message`](Payload::set_message). For the bytes the
/// dispatch path will actually send downstream, use
/// [`encoded()`](Payload::encoded).
pub fn bytes(&self) -> &Bytes {
```

If renaming is on the table while it's pre-1.0, `received_bytes()` or `original_bytes()` would resolve the ambiguity at the call site rather than the docstring.

### M2 — `message()` doc is silent on which error code each failure mode produces

The `# Errors` section says "Returns an error if the bytes fail to decode, if a replacement … is not an `M`, or if a previous `message::<N>()` cached a different type than `M`." Three failure modes, two error codes (`invalid_argument` for decode failure, `internal` for both type-mismatch cases) — and the distinction is *intentional and correct* (peer-supplied bytes vs. server-side type bug). The doc should say so:

```rust
/// # Errors
///
/// - [`invalid_argument`](ConnectError::invalid_argument) if the wire
///   bytes fail to decode as `M` — peer-supplied data, not a server bug.
/// - [`internal`](ConnectError::internal) if a replacement set with
///   [`set_message`](Payload::set_message) is not an `M`, or if a prior
///   `message::<N>()` cached a different type — both are server-side
///   programming errors (interceptors and handlers for the same RPC must
///   agree on the message types).
```

### M3 — Test gaps

All cheap to add and they'd pin behaviors the doc promises:

- `message()` on corrupt wire bytes (the `invalid_argument` decode-error path — distinct from the wrong-type path that `message_wrong_type_errors` covers).
- The replacement wrong-type path (`payload.rs:162–169`) — only the post-decode wrong-type path is tested, and the error message is different (`"payload replacement is a {}, not a {}"` vs `"payload was previously decoded as a {}, not a {}"`).
- `view()` on a *JSON-format* payload that has been *replaced* via `set_message` (the cross-format proto re-encode at `payload.rs:218`).
- `set_message` called twice (second supersedes first).
- A concurrent `message::<M>()` / `message::<N>()` race — the comment at `payload.rs:171–176` describes the semantics precisely; a test would pin them.

### M4 — `set_message` reset polish

`payload.rs:242–245` resets `decoded` to a fresh `OnceLock`. It's not needed for correctness (`replaced` is checked first), and when nothing was decoded yet it allocates a fresh `OnceLock` just to drop an empty one. The intent (drop the prior box's memory) is fine; a guard makes it precise and cheap:

```rust
self.replaced = Some(Box::new(message));
// Drop the prior decode cache so the original message doesn't pin
// memory for the Payload's lifetime. `replaced` is checked first, so
// a stale cache would never be visible — this is purely a memory
// concern.
if self.decoded.get().is_some() {
    self.decoded = OnceLock::new();
}
```

## Low / for the series, not this PR

- The `AnyMessage` round-trip invariant doc (`payload.rs:45–48`) could note that violating it causes `encoded()` to silently produce wrong-shape bytes — no panic, no error, just corrupt wire output. That's the only place a manual-impl violation surfaces.
- An interceptor that calls `message()` decodes once; the handler then re-decodes the same bytes (`call_unary` takes `Bytes`, not `Payload`). Fine for the foundation PR — worth a note somewhere so it's tracked as a future optimization (the handler dispatch could take `Payload` and reuse the cache).

iainmcgin (approved):

PR #114, opened 2026-05-18
interceptor: Interceptor trait, Next continuation, server registration

aknott-ant:
> **Stacked on #113** (`aknott/interceptors-pr2-payload`), which is stacked
> on #112.
>
> This PR targets `main`, so its diff currently also includes the #112 and
> #113 commits. **Review only the top commit** — _"interceptor: Interceptor
> trait, Next continuation, server registration"_. Once #112 and #113 merge,
> this branch will be rebased and the diff will reduce to just that commit.
> Kept as a draft until #112 and #113 land.

## Summary

PR 3 of the interceptor series — the user-facing surface. Adds a
connect-go-style typed unary interceptor. Interceptors wrap a single RPC
after envelope decoding, decompression, and header parsing, and before the
handler — they see the parsed `Spec`, headers, deadline, extensions, and a
lazily-decoded `Payload`, and can rewrite the request and response messages.

- **`Interceptor`** async trait with a default-passthrough `intercept_unary`.
  Annotate impls with the re-exported `#[connectrpc::async_trait]` — there is
  no separate `async-trait` dependency for downstream crates.
- **`Next<'a>`** continuation: holds the rest of the chain and a
  `pub(crate)` terminal. `run(self, req)` consumes it; an interceptor can
  call it at most once (enforced by the type system), and not calling it
  short-circuits the chain.
- **`UnaryRequest { ctx, payload }`** (`#[non_exhaustive]`) and
  **`UnaryResponse = Response<Payload>`** so `with_header` / `with_trailer` /
  `compress` / `map_body` are inherited.
- **`unary_interceptor(closure)`** for closure-shaped interceptors, and
  `interceptor::run_chain(chain, req, terminal_fn)` so interceptor authors
  can unit-test without a tower service or TCP listener.
- **`ConnectRpcService::with_interceptor`** builder; chain stored as
  `Arc<[Arc<dyn Interceptor>]>` for one Arc bump per request.
- **`call_unary_intercepted`** with an `is_empty()` fast path that delegates
  straight to `dispatcher.call_unary` — when no interceptors are registered
  the dispatch path is byte-identical to before.

Threaded through the five `call_unary` sites in `service.rs` (Connect unary,
Connect GET, gRPC fast-path, gRPC streaming-frame unary, and the generic
fallback). Streaming RPCs bypass the chain in this release.

## Interceptor series

1. #112 — `Spec` / `StreamType` / `IdempotencyLevel`
2. #113 — `AnyMessage` / `Payload`
3. **(this PR)** — `Interceptor` trait, `Next`, server registration

## Verification

Run on the rebased branch (buffa 0.6.0):

- `task generate:all` — no generated-code drift
- `task lint` — pass
- `task test` — pass, 0 failed
- Conformance: 3600/3600 (the suite runs with zero interceptors)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ tokio-stream = "0.1"
 tokio-util = "0.7"
 futures = "0.3"
 pin-project = "1"
+async-trait = "0.1"
 
 # Protobuf. The 0.6 floor matches the regen baseline for the checked-in
 # generated code: 0.6.0 adds `with_*` setters for explicit-presence fields,
--- a/connectrpc/Cargo.toml
+++ b/connectrpc/Cargo.toml
@@ -28,6 +28,7 @@ tower = { workspace = true }
 tokio = { workspace = true, features = ["rt", "io-util", "sync", "time"] }
 futures = { workspace = true }
 pin-project = { workspace = true }
+async-trait = { workspace = true }
 
 # Protobuf
 buffa = { workspace = true }
--- a/connectrpc/src/interceptor.rs
+++ b/connectrpc/src/interceptor.rs
@@ -1,0 +1,640 @@
+//! RPC-level interceptors for unary calls.
+//!
+//! Interceptors are the typed equivalent of `tower` middleware: they wrap
+//! a single RPC *after* envelope decoding, decompression, and header
+//! parsing, and *before* the handler runs. Each interceptor sees a
+//! [`UnaryRequest`] (the [`Spec`](crate::Spec), headers, deadline,
+//! extensions, and a lazily-decoded [`Payload`]) and a [`Next`]
+//! continuation, and returns a [`UnaryResponse`].
+//!
+//! The first interceptor registered is the outermost: it runs first on
+//! the way in and last on the way out, exactly like wrapping a function
+//! call. This matches `connect-go`'s `WithInterceptors` ordering.
+//!
+//! ```text
+//! request ──▶ interceptor[0] ──▶ interceptor[1] ──▶ handler
+//!                  │                  │                 │
+//! response ◀───────┴──────────◀───────┴────────◀───────┘
+//! ```
+//!
+//! Register interceptors with
+//! [`ConnectRpcService::with_interceptor`](crate::ConnectRpcService::with_interceptor).
+//! When no interceptors are registered the dispatch path is byte-for-byte
+//! identical to a build without this module — there is no per-request
+//! cost for opting out.
+
+use std::sync::Arc;
+
+use bytes::Bytes;
+use futures::future::BoxFuture;
+
+use crate::codec::CodecFormat;
+use crate::error::ConnectError;
+use crate::payload::Payload;
+use crate::response::{EncodedResponse, RequestContext, Response};
+
+/// Re-export of [`async_trait::async_trait`] so interceptor authors don't
+/// need a direct `async-trait` dependency.
+///
+/// ```rust,ignore
+/// #[connectrpc::async_trait]
+/// impl connectrpc::Interceptor for MyInterceptor { /* ... */ }
+/// ```
+///
+/// The macro expansion references only `core` and the prelude — there is
+/// no runtime `async-trait` requirement.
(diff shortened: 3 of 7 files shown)


aknott-ant:

iainmcgin:
[claude code] Reviewed after rebasing onto `main` (post-#113/#116/#117). Clean rebase; one mechanical fixup folded into your commit (the `path_flows_to_handler_context` test from #116 needed the new `&[]` interceptors arg). 327 tests pass.

The design is sound. `Next` consuming `self` so it can be called *at most once* is a real improvement over connect-go's `next func(...)` (which is callable twice and double-runs the chain). The `Arc<[Arc<dyn Interceptor>]>` storage and the `is_empty()` fast path that skips `UnaryRequest`/`Payload` construction entirely are exactly right for the no-interceptor common case. The `interceptor_wraps_unary_call` e2e test exercising the full Tower + axum stack is good.

Two things I'd address before merge, both small.

## High

### H1 — The trait doc example canonicalizes the pattern `path()` exists to replace

`interceptor.rs:76`:

```rust
let proc = req.ctx.spec().map(|s| s.procedure).unwrap_or("<dynamic>");
```

Now that `RequestContext::path()` (#116) is in `main`, this is the wrong example to lead with. `spec()` is `None` for dynamic `Router` dispatch, and falling back to `"<dynamic>"` means an auth interceptor or span builder copying this pattern silently loses the method name for an entire class of routes — the exact gap `path()` was added to close.

The trait doc is what every interceptor author starts from. It should read:

```rust
// `ctx.path()` is the requested procedure path, populated unconditionally
// by the dispatch path. `ctx.spec()` is the *resolved* method's static
// metadata and is `None` for the dynamic `Router`; use it for
// `stream_type`/`idempotency`, not for the method name.
let proc = req.ctx.path().unwrap_or("<unknown>");
tracing::info!(%proc, "rpc start");
```

(The `unwrap_or` is for hand-built test contexts; the dispatch path always sets `path()` before the interceptor runs.)

## Medium

### M1 — No test pins that a short-circuiting interceptor's `ConnectError` headers reach the wire

`short_circuit_skips_terminal` asserts `err.code` but not response headers. An auth interceptor that returns `Err(ConnectError::permission_denied("...").with_headers(diagnostic_headers))` relies on those headers landing in the wire response. The error renderers (`error_response`, `grpc_error_response`) do walk `response_headers()` — but a regression there would land silently because nothing exercises the interceptor → error → wire path end-to-end. Worth one assertion in the existing test, or one new e2e test in `tests/streaming/`.

### M2 — Interceptor coverage is one-protocol-shape

The commit message says interceptors are threaded through "Connect unary, Connect GET, gRPC fast-path, gRPC streaming-frame unary, and the generic fallback." The only test (`interceptor_wraps_unary_call`) is Connect POST. The other four shapes are wired identically (`call_unary_intercepted` at each site), so a regression *probably* hits them all together — but the Connect GET path in particular is the one that matters for an authz interceptor: an idempotent unary RPC that opts into HTTP GET must not bypass the chain. One test that registers an interceptor and makes a GET request, asserting it ran, would close the gap. The gRPC fast-path one is cheaper to add inline alongside `empty_chain_is_no_op` since the test infra is already there.

## Low / not for this PR

- `with_interceptor` always wraps in a fresh `Arc`. There's no way to register an existing `Arc<dyn Interceptor>` (sharing one stateful interceptor across multiple `ConnectRpcService`s). The workaround is fine — interceptors carry their shared state internally as `Arc<...>`, so the *interceptor* doesn't need to be shared — but a `with_interceptor_arc` overload is a cheap ergonomic add if anyone asks.
- `DispatchTerminal::call` does `payload.encoded()` (one `Bytes` clone for the no-replacement case) and the handler's `call_unary` decodes again. An interceptor that calls `payload.message()` decodes once; the handler decodes a second time on the same bytes. Fine for the foundation PR, worth a tracking note for a future change that has the dispatcher accept `Payload` directly.

---

I'll push the H1 fix and the M1/M2 tests directly per Iain's go-ahead.

iainmcgin (approved):

PR #117, opened 2026-05-19
client: HttpClientBuilder with connect_timeout

azdagron:
## Motivation

`HttpClient::plaintext()` / `plaintext_http2_only()` / `with_tls()` each construct an `HttpConnector` inline with no way to set `connect_timeout`. The TCP connect phase is therefore bounded only by the kernel's `tcp_syn_retries` (~130s on Linux defaults). Callers on network paths that can silently drop SYNs — NAT gateways under churn are the motivating case — have no way to fail fast short of a per-request timeout, which conflates the connect phase with TLS handshake and request body.

## What

Adds `HttpClientBuilder`, obtained via `HttpClient::builder()`, with a `connect_timeout(Duration)` setter that flows through to `HttpConnector::set_connect_timeout` for all three transport flavours. The three existing associated constructors now delegate to the builder with no timeout set, so behaviour is unchanged for existing callers and the connector setup lives in one place.

The setter bounds only the TCP `connect(2)` call (per resolved address — hyper divides the budget across the address set). It does **not** cover DNS resolution or, for `with_tls`, the TLS handshake. The doc comment points callers at `CallOptions::with_timeout` for an end-to-end bound; `connect_timeout` is for the case where you want to retry connect quickly without aborting an in-flight request body.

## Not in this PR

`Http2Connection` / `MakeSendRequest` in `client/http2.rs` build their own `HttpConnector` instances and are left untouched. Happy to follow up with the same builder shape there if there's demand; the immediate consumer uses `HttpClient`.

## Testing

`http_client_connect_timeout_bounds_tcp_connect` sends to a TEST-NET-1 address (RFC 5737, guaranteed unroutable) with a 100 ms connect timeout and asserts the failure surfaces within 2 s — well under the kernel SYN-retry floor an unbounded connect would hit. Existing 324 lib tests pass unchanged.

### DIFF
--- a/connectrpc/src/client/mod.rs
+++ b/connectrpc/src/client/mod.rs
@@ -314,6 +314,16 @@ enum HttpClientInner {
 #[cfg(feature = "client")]
 #[cfg_attr(docsrs, doc(cfg(feature = "client")))]
 impl HttpClient {
+    /// Returns a builder for configuring connector-level options before
+    /// choosing a transport flavour.
+    ///
+    /// `HttpClient::plaintext()` is equivalent to
+    /// `HttpClient::builder().plaintext()`, and likewise for the other
+    /// constructors.
+    pub fn builder() -> HttpClientBuilder {
+        HttpClientBuilder::default()
+    }
+
     /// Create a **plaintext** HTTP client. Only for `http://` URIs.
     ///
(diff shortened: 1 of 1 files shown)


azdagron:

iainmcgin (approved):

PR #119, opened 2026-05-19
dispatcher: pass Payload to call_unary so handlers reuse the interceptor decode

iainmcgin:
## Why

When an interceptor reads the request body via `Payload::message::<M>()`, the decode is cached in `Payload`'s `OnceLock`. `DispatchTerminal::call` then stripped the `Payload` back to raw `Bytes` (`payload.encoded()`) and handed them to `Dispatcher::call_unary`, which decoded the same bytes a second time for the handler. The interceptor's cache was discarded — every body-inspecting interceptor pays for the decode twice.

## What

Pass the `Payload` through the dispatcher contract so the handler's decode can reuse the interceptor's cache.

- **`Dispatcher::call_unary` takes `Payload`, not `Bytes`.** `ErasedHandler` and the `Router`/codegen impls follow. Pre-1.0 breaking change.
- **`Payload::take_message::<M>()`** consumes `self` and returns an owned `M`: the cached decode if one exists (no re-decode, the `Box<M>` is unboxed in place), the replacement if `set_message` was called, or a fresh decode otherwise. Owned-message handlers (`Router::route`) call this.
- **`AnyMessage::into_any(self: Box<Self>) -> Box<dyn Any>`** so a cached `Box<dyn AnyMessage>` can be downcast and moved out by value. Breaking for manual `AnyMessage` impls (the doc says "you will almost never" write one); the blanket impl covers every codegen message.
- **Generated dispatchers and zero-copy view handlers call `Payload::encoded()`** — the cache stores owned messages, not views, so it cannot help the view path. `encoded()` is a refcounted `Bytes` clone for the common no-replacement case, the same cost as before.
- **The empty-interceptor-chain fast path keeps its no-overhead invariant.** `Payload::new` is plain struct construction (no allocation, no decode).

### Behavioral change worth noting

Before this PR, an interceptor that decoded the body as the *wrong* message type failed silently: the handler decoded the wire bytes independently and never saw the interceptor's mistake. Now `Payload::take_message` surfaces the wrong-type cache as an `Internal` error at the handler. This is the intended behavior — interceptors and handlers for the same RPC must agree on the message types — but it's a change. The `take_message_wrong_cached_type_errors` test pins it.

## Confidence

- 6 new `Payload::take_message` unit tests (fresh decode, cache reuse, replacement, wrong cache type, wrong replacement type, decode error code).
- New `dispatch_terminal_forwards_payload_to_handler` interceptor test: hands the dispatcher a `Payload` whose wire bytes are garbage but whose cache holds a replacement — if the terminal stripped the `Payload` to bytes (the old behavior), the dispatcher's decode would error; if it forwards the `Payload`, the dispatcher sees the replacement.
- 336 `connectrpc` lib tests, 20 `streaming-test` e2e tests pass. `cargo clippy --workspace --all-targets -- -D warnings` clean. `cargo fmt` clean. `cargo doc` introduces no new warnings.
- `task generate:all` regenerated all checked-in `*.__connect.rs` files; the diff there is just the `request: Payload` signature change and the `request.encoded()?` call.

### DIFF
--- a/benches/rpc/src/generated/connect/bench.__connect.rs
+++ b/benches/rpc/src/generated/connect/bench.__connect.rs
@@ -490,7 +490,7 @@ impl<T: BenchService> ::connectrpc::Dispatcher for BenchServiceServer<T> {
         &self,
         path: &str,
         ctx: ::connectrpc::RequestContext,
-        request: ::buffa::bytes::Bytes,
+        request: ::connectrpc::Payload,
         format: ::connectrpc::CodecFormat,
     ) -> ::connectrpc::dispatcher::codegen::UnaryResult {
         let Some(method) = path.strip_prefix("bench.v1.BenchService/") else {
@@ -503,7 +503,7 @@ impl<T: BenchService> ::connectrpc::Dispatcher for BenchServiceServer<T> {
                 Box::pin(async move {
                     let req = ::connectrpc::dispatcher::codegen::decode_request_view::<
                         crate::proto::bench::v1::__buffa::view::BenchRequestView,
-                    >(request, format)?;
+                    >(request.encoded()?, format)?;
                     svc.unary(ctx, req)
                         .await?
                         .encode::<crate::proto::bench::v1::BenchResponse>(format)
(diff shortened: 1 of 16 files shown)


aknott-ant (approved):
:shipit:

PR #124, opened 2026-05-20
examples/bazel: regenerate Cargo.lock against connectrpc 0.6.0 / buffa 0.6.0

iainmcgin:
## Why

The standalone bazel example's `Cargo.lock` had been pinned at `connectrpc 0.3.1` / `buffa 0.3.0` since the example landed. `Cargo.toml` was bumped to 0.5.0 and then 0.6.0 in the release PRs, but the lock was never regenerated — it's a non-workspace manifest (`[workspace]` table is empty), so workspace-wide `cargo update` doesn't reach it, and the example isn't built in CI.

`crates_universe` reads this `Cargo.lock` via `crate.from_cargo(cargo_lockfile = "//:Cargo.lock")`, so a stale lock means a `bazel build` either fails or quietly pulls 0.3.x.

## What

`cargo update` against the now-published `connectrpc 0.6.0` and `buffa 0.6.0` crates. Transitive deps follow (tokio 1.50 → 1.52, etc.). `MODULE.bazel.lock` is gitignored and regenerates on the next `bazel build`.

## Verified

`cargo metadata` resolves cleanly with `connectrpc 0.6.0` / `buffa 0.6.0`. The bazel example is not in CI; full `bazel test //...` requires a sibling `~/src/buffa` checkout for `setup.sh`, which isn't part of the verification here.

### DIFF
--- a/examples/bazel/Cargo.lock
+++ b/examples/bazel/Cargo.lock
@@ -3,6 +3,17 @@
 version = 4
 
 [[package]]
+name = "async-trait"
+version = "0.1.89"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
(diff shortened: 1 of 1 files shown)


rpb-ant (approved):
:shipit:

PR #126, opened 2026-05-20
ci: check generated code freshness

### LINKED ISSUE #95: ci: add check-generated-code guard for checked-in generated Rust
## Problem

connect-rust has checked-in generated Rust in four places, regenerated by `task generate:all`:

- `examples/eliza/src/generated/`
- `examples/multiservice/src/generated/`
- `conformance/src/generated/`
- `benches/rpc/src/generated/`

CI doesn't verify those are up to date. A codegen change to `protoc-gen-connect-rust`, `connectrpc-codegen`, or the bundled `protoc-gen-buffa`/`protoc-gen-buffa-packaging` toolchain can merge without `task generate:all` having been re-run, leaving the checked-in code stale until someone notices the diff in a later, unrelated PR.

This matters more than usual here because the examples are documentation — `examples/eliza` is what a downstream consumer copies to get started, and `conformance/src/generated/` is what the conformance suite actually exercises. Stale generated code means the examples no longer demonstrate what the current toolchain produces, and the conformance suite isn't testing the current codegen.

Concrete motivating case: anthropics/buffa#102 changes `buffa_codegen::ALLOW_LINTS` (adds `unused_qualifications`), which changes the `#[allow(...)]` block `protoc-gen-buffa-packaging` writes into every generated `mod.rs`. In the buffa repo, a `check-generated-code` CI job caught the stale checked-in code immediately. In connect-rust, the same change would land via the next `protoc-gen-buffa` bump and nothing would notice that `examples/eliza/src/generated/buffa/mod.rs` (and every other `mod.rs` in the four trees) no longer matches what the toolchain produces.

## Proposal

Add a `check-generated-code` job to CI that mirrors `task generate:all` and fails on any diff in the four checked-in trees, modeled on [buffa's `check-generated-code` job](https://github.com/anthropics/buffa/blob/main/.github/workflows/ci.yml).

```yaml
check-generated-code:
  runs-on: ubuntu-latest
  timeout-minutes: 15
  steps:
    - uses: actions/checkout@v4
    - uses: actions/checkout@v4
      with:
        repository: anthropics/buffa
        path: buffa-sibling
        # Pin to the buffa version in Cargo.lock so the sibling toolchain
        # matches what consumers would resolve from the registry.
        ref: v0.5.1
    - uses: dtolnay/rust-toolchain@stable
    - uses: Swatinem/rust-cache@v2
    - uses: bufbuild/buf-action@v1
      with: { setup_only: true }
    - name: Build codegen plugins
      run: |
        cargo build --release -p connectrpc-codegen --bin protoc-gen-connect-rust
        cargo build --release --manifest-path buffa-sibling/Cargo.toml \
          -p protoc-gen-buffa -p protoc-gen-buffa-packaging
    - name: Regenerate
      # The Taskfile expects ../buffa; symlink or override the path var.
      run: |
        ln -sfn "$PWD/buffa-sibling" ../buffa
        task generate:all
    - name: Check for differences
      run: |
        if ! git diff --exit-code examples/*/src/generated/ conformance/src/generated/ benches/rpc/src/generated/; then
          echo "::error::Checked-in generated code is stale. Run 'task generate:all' and commit."
          exit 1
        fi
```

## Open questions

- **Where to source the sibling `buffa` toolchain.** `task generate:all` builds `protoc-gen-buffa`/`protoc-gen-buffa-packaging` from `../buffa`, which CI doesn't have. Options: (a) clone `anthropics/buffa` at the tag matching `Cargo.lock` (sketched above), (b) download the pre-built plugin binaries from the buffa GitHub release matching the locked version, (c) `cargo install protoc-gen-buffa protoc-gen-buffa-packaging --version <locked>` from crates.io. (b) is fastest and avoids a sibling-repo build; (a) is most robust to a buffa version not yet on the release page.
- **Whether `benches/rpc/src/generated/` should be in scope.** It's checked in but only used for benchmarks; staleness is lower-risk than the examples and conformance trees. Cheap to include either way.
- **Pin discipline.** The job should resolve the buffa toolchain version from `Cargo.lock` rather than hardcode a tag, so a buffa dep bump and the regen land atomically.

Yong-yuan-X:
## Summary

- Add a `check-generated-code` CI job that runs `task generate:all`
- Resolve the sibling `buffa` checkout from the version locked in `Cargo.lock`
- Fail CI when checked-in generated code is stale
- Update contributing docs to mention the generated-code CI check

Closes #95.

## Verification

Environment:

- `task 3.50.0`
- `buf 1.69.0`
- `protoc libprotoc 33.5`
- sibling `../buffa` checked out at `v0.6.0`

Commands:

- `task generate:all`
- `git diff --exit-code -- conformance/src/generated examples/eliza/src/generated examples/multiservice/src/generated benches/rpc/src/generated`
- `git diff --check`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,6 +89,12 @@ jobs:
     name: Check generated code
     runs-on: ubuntu-latest
     timeout-minutes: 15
+    # Note: the workspace has no committed Cargo.lock, so `cargo
+    # generate-lockfile` below resolves buffa and prettyplease (which formats
+    # the generated code) to the latest compatible releases at run time. A new
+    # release of either can therefore make this job fail on PRs that did not
+    # touch the protos or codegen — the fix is to run `task generate:all` on
+    # main and commit the resulting diff.
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
@@ -100,6 +106,10 @@ jobs:
       - uses: bufbuild/buf-action@fd21066df7214747548607aaa45548ba2b9bc1ff # v1 (sha-pinned)
         with:
           setup_only: true
+          # Pinned so the job's behaviour does not drift with new buf
+          # releases; keep in sync with the version used for local
+          # regeneration when bumping.
+          version: 1.69.0
       - uses: go-task/setup-task@01a4adf9db2d14c1de7a560f09170b6e0df736aa # v2 (sha-pinned)
       - name: Generate lockfile
         run: cargo generate-lockfile
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,6 +85,64 @@ jobs:
           components: rustfmt
       - run: cargo +nightly fmt --all -- --check
 
+  check-generated-code:
+    name: Check generated code
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1 (sha-pinned)
+      - uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b # v3.0.0 (sha-pinned)
+        with:
+          version: ${{ env.PROTOC_VERSION }}
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - uses: bufbuild/buf-action@fd21066df7214747548607aaa45548ba2b9bc1ff # v1 (sha-pinned)
+        with:
+          setup_only: true
+      - uses: go-task/setup-task@01a4adf9db2d14c1de7a560f09170b6e0df736aa # v2 (sha-pinned)
+      - name: Generate lockfile
+        run: cargo generate-lockfile
+      - name: Resolve locked buffa version
+        id: buffa-version
+        shell: bash
+        run: |
+          version="$(awk '
+            $0 == "[[package]]" { in_buffa = 0 }
+            $0 == "name = \"buffa\"" { in_buffa = 1 }
+            in_buffa && $1 == "version" {
+              gsub(/"/, "", $3)
+              print $3
+              exit
+            }
+          ' Cargo.lock)"
+
+          if [ -z "$version" ]; then
+            echo "::error::Could not find the buffa package version in Cargo.lock"
+            exit 1
+          fi
+
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+      - uses: actions/checkout@v4
+        with:
+          repository: anthropics/buffa
+          ref: v${{ steps.buffa-version.outputs.version }}
+          path: buffa
+      - name: Link buffa sibling checkout
+        shell: bash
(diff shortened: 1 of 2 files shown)


Yong-yuan-X:
Looks like the generated-code check failed because `Cargo.lock` is not tracked in this repo, but the workflow tried to read it before generating one.I added `cargo generate-lockfile` before resolving the locked `buffa` version, and rewrote the branch into one signed commit for the verified-signatures requirement.❤

iainmcgin (commented):
[claude code] Thanks — this closes a real gap (#95): today stale generated code only surfaces indirectly as compile errors in `Check`/`Test`, and a dedicated job with a clear "run `task generate:all` and commit the diff" message is a much better failure mode. The action SHA-pinning and the `${{ env.PROTOC_VERSION }}` reuse are consistent with how the existing jobs are set up, and the CONTRIBUTING.md updates match the new behavior. A few points before merging:

**1. Please pin the `buf` version in `bufbuild/buf-action`.** With `setup_only: true` and no `version`, the job installs whatever the latest buf release is on the day it runs, so the job's behavior can drift over time independent of anything in the repo. The other toolchain inputs in this workflow are pinned (protoc via `PROTOC_VERSION`, actions by SHA); buf should follow suit — a `version:` input plus a comment noting it's the same buf used for local regeneration is enough.

**2. Known flake mode worth a comment in the job.** The workspace has no committed `Cargo.lock`, so `cargo generate-lockfile` resolves buffa — and, more importantly, `prettyplease`, which formats all generated output — to the latest compatible versions at run time. A new prettyplease (or buffa) release can therefore turn this job red on PRs that touched nothing, until someone regenerates and commits the diff. That's arguably the job doing its job (a maintainer regenerating locally would see the same diff), but it's worth a short comment in the workflow so whoever sees the failure understands why it can happen without any change to the protos or codegen. If you'd rather avoid that mode entirely we can discuss pinning options, but I don't think it needs to block this PR.

**3. Optional simplification:** the awk over `Cargo.lock` works, but `cargo pkgid buffa` (after `cargo generate-lockfile`) prints `…buffa@<version>` and is less sensitive to lockfile formatting; `version="${$(cargo pkgid buffa)##*@}"` would replace the whole block. Take it or leave it.

**4. Timeout:** `task generate:all` builds `protoc-gen-connect-rust` and the buffa plugins in release mode; on a cold cache that plus the checkout/setup steps may brush up against `timeout-minutes: 15`. If the first real run times out, bumping to 20–25 is fine — no need to change preemptively.

The job hasn't executed in CI yet (the workflow run needs maintainer approval, which we'll sort out on our side) — we'd want to see it green here before merging, since the interesting failure modes (sibling `../buffa` layout, plugin builds, buf invocation) are exactly the parts that only show up in the runner environment. No action needed from you on that beyond keeping the branch as-is; we'll kick the run and follow up if anything in the job needs adjusting.

iainmcgin:
[claude code] To save a round trip I've pushed a small commit to this branch (maintainer edit) applying points 1 and 2 from the review: the buf release installed by `buf-action` is now pinned (1.69.0, matching the version you verified with), and the job has a header comment explaining the fresh-resolution failure mode so future readers understand why it can go red without a proto/codegen change. The awk-based version resolution is left as you wrote it — point 3 was optional. Next step on our side is approving the workflow run so the job can prove itself in CI.

iainmcgin (approved):

PR #128, opened 2026-05-21
Add new `connectrpc-health` crate

torkelrogstad:
Inspired by [connectrpc.com/grpchealth](https://pkg.go.dev/connectrpc.com/grpchealth) and the equivalent Tonic implementation. Breaking up into four different commits felt kind of awkward, but did it per CONTRIBUTING.md guidelines.


torkelrogstad:

iainmcgin (commented):
[claude code] Thanks — this is a strong contribution and we'd like to take it. We ran it through both a correctness/concurrency review and an API-ergonomics review; the hand-written code came out of both in very good shape: locks are block-scoped and never held across awaits, `send_if_modified` suppresses no-op wake-ups, dropping a Watch stream releases the subscriber (with a regression test for exactly that), mutex poisoning is recovered rather than propagated, peer-supplied service names are only ever used as lookup keys (no insertion path a caller can drive), and every documented behaviour of `StaticChecker` has a colocated test. The deliberate deviation on `Watch` for unknown services (NOT_FOUND rather than the spec's open stream with SERVICE_UNKNOWN, matching connect-go's grpchealth) is documented consistently in all four places it matters — that's fine by us. The buf/Taskfile wiring matches the other generated directories and the generated output follows the fully-qualified-paths convention.

The asks below are about the public API surface, since this becomes a published crate.

**Before merge**

1. **Feature-gate the client half.** `connectrpc-health/Cargo.toml:16` unconditionally enables `connectrpc`'s `server` *and* `client` features, so a server that only exposes the health service still pulls in the whole client transport stack. Please add a `client` feature on this crate (off by default or default — your call) that forwards to `connectrpc/client`, and gate the `HealthClient`/`HealthExt` re-exports behind it.
2. **Make the common case one line and leave the user holding a usable handle.** Today the static, always-SERVING case is two `Arc::new`s, an `Arc::clone`, `from_arc`, and `register` — and the discoverable constructor (`HealthService::new`) seals the checker inside the service so there's nothing left to call `set_status` on. A crate-level convenience along the lines of `install_static(router, [SERVICE_NAME]) -> (Router, Arc<StaticChecker>)` (keeping `from_arc` for custom checkers) would close most of the gap to grpchealth/tonic-health. Related: `checker(&self) -> &Arc<C>` is an unusual return type — either return `Arc<C>` (clone internally) or have the doc spell out `Arc::clone(svc.checker())`, since "Clone it" currently steers users into `clippy::clone_on_ref_ptr` territory.
3. **Decide and document the `set_status` register-on-miss semantics.** `StaticChecker::set_status` silently creates an entry for any name (`static_checker.rs:61-76`), so a typo'd service name "works" while probes for the real FQN keep reporting the old status, and the map grows if the name ever comes from non-static input. connect-go's `StaticChecker.SetStatus` errors on unknown names. Either split registration from updates (or add a strict `try_set_status`), or keep the current behaviour but say loudly on `set_status` itself that it registers on miss and must only be fed the generated `*_SERVICE_NAME` constants; a `remove_service` would round the API out either way.
4. **Guide fixes.** The example at `docs/guide.md:1001` doesn't compile: `Status` is used but not imported, and the Cargo snippet above it omits the `connectrpc = { version = "0.6", features = ["server"] }` dependency the example needs. Marking the block `rust,no_run` (like the lib.rs Quick Start) would let CI catch future drift.
5. **Crate metadata.** Please add a `readme` (the sibling crates point at the workspace README) and bump the crate version to match the workspace now that 0.6.1 is out.

**Fine as follow-ups (or in this PR if you prefer)**

- `From<ServingStatus> for Status` (or `TryFrom`) for probe-loop code decoding raw responses; the forward direction already exists.
- The default `Checker::watch` returning `Unimplemented` is spec-legal, but it's easy for a custom-`Checker` user to ship without noticing Watch is dead — consider a doc callout right on the trait method (or a `tracing` warning) rather than only under `# Errors`.
- A `with_services_status(...)` form so a service can be registered as NotServing from the start.
- The `#[allow(clippy::upper_case_acronyms)]` on the generated-module wrappers appears unused.
- An end-to-end test that a client disconnect mid-Watch drops the server-side subscriber (the WatchStream-drop unit test is convincing; the hyper-level version would tie it together).

On process: CI for first-time contributors needs a maintainer to approve the workflow run — we'll do that once the above lands, and the new "Check generated code" job that just merged should pass given the Taskfile/buf wiring you've already included. Happy to re-review quickly after the changes.

torkelrogstad:
Thanks for your review on this @iainmcgin. I've added a new commit at the start that modifies the codegen to achieves the feature-gate you requested. The other commits have been amended to incorporate your feedback. 

While working on the codegen changes I realized it would make sense to expose codegen knobs for disabling client/server generation all together, if you're only interested in one of them. Skipped this here, as that seemed like a bigger change. But curious what you think about that.

torkelrogstad:
@iainmcgin friendly ping

iainmcgin:
> @iainmcgin friendly ping

apologies, I was out on vacation last week - I'm back now, and I'll get this in.

iainmcgin (approved):
[claude code] Re-reviewed the full series — everything we asked for landed, and landed well.

- The `client` feature gate is exactly the dep-forwarding pattern we hoped for, and doing it properly via an opt-in `gate_client_feature` codegen option (default-off, so external consumers' output is byte-identical) was the right call. The `no_ungated_client_references` syn-walker test is a genuinely strong guard — thanks for going beyond the count assertion.
- `install_static` closes the ergonomics gap, `checker()` now returns `Arc<C>`, and `set_status` erroring on unknown names (with `register`/`register_with_status`/`remove_service` rounding it out) matches grpchealth semantics.
- You also picked up most of the "fine as follow-ups" list — `From<ServingStatus>`, the Watch doc callout on the trait method, `register_with_status`, and the end-to-end disconnect test. Appreciated.

We re-ran both review passes (correctness/concurrency and API ergonomics) over the revision: no blocking findings. Two cosmetic notes we'll handle ourselves in a follow-up rather than ask for another round: the guide example uses rustdoc hidden-line `#` syntax, but `docs/guide.md` is rendered as plain markdown (it isn't doctested), so those lines show up literally on GitHub; and we may drop the `#[must_use]` on `register`/`register_with_status` since discarding that bool is the normal case.

On your codegen question — yes, we'd take `disable_client` / `disable_server` generation knobs as a separate PR. Suppressing emission entirely is a nice complement to cfg-gating (no feature declaration needed in the consumer, smaller generated files), and the option-threading you added here is most of the plumbing. Feel free to open an issue or PR for it.

CI is green across all 12 jobs, including the new generated-code freshness check and both conformance suites. Merging — thanks for a model contribution.

PR #131, opened 2026-05-27
compression: error on truncated gzip deflate streams

iainmcgin:
## Summary

`GzipProvider::decompress_inner` loops until the decompressor reports end-of-stream, but a deflate stream that ends without an end-of-stream marker never reaches that state: once the input is exhausted, `decompress_vec` keeps returning `Status::BufError` and the loop makes no further progress. Since output capacity is always ensured before each call, `BufError` can only mean the remaining input cannot complete the stream, so it is now treated as a hard decompression error ("truncated or invalid deflate stream").

Behavior for well-formed gzip data is unchanged; the existing roundtrip, pooled-decompressor, and cross-compatibility tests all pass.

## Tests

Three new provider-level tests: a header-only gzip member (no deflate data, no trailer), a stream truncated in the middle of the deflate data, and a stream missing its 8-byte CRC/length trailer.

### DIFF
--- a/connectrpc/src/compression.rs
+++ b/connectrpc/src/compression.rs
@@ -792,7 +792,7 @@ impl GzipProvider {
                 // loop would never terminate on such input.
                 flate2::Status::BufError => {
                     return Err(ConnectError::internal(
-                        "gzip decompression failed: truncated or invalid deflate stream",
+                        "gzip decompression stalled: truncated or invalid deflate stream",
                     ));
                 }
             }
@@ -1425,15 +1425,20 @@ mod tests {
     #[cfg(feature = "gzip")]
     const TRUNCATION_TEST_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
 
+    /// A minimal, valid gzip member header with nothing after it:
+    /// id1, id2, CM=8 (deflate), FLG=0, MTIME=0, XFL=0, OS=0xff (unknown).
+    #[cfg(feature = "gzip")]
+    const MINIMAL_GZIP_HEADER: [u8; 10] =
+        [0x1f, 0x8b, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xff];
+
     /// A gzip member that is only a header — no deflate data, no trailer —
     /// must be rejected rather than treated as an incomplete stream to wait
     /// on.
     #[cfg(feature = "gzip")]
     #[test]
     fn test_gzip_decompress_header_only() {
-        let header_only = [0x1f, 0x8b, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xff];
         let err = run_with_timeout(TRUNCATION_TEST_TIMEOUT, move || {
-            GzipProvider::default().decompress_with_limit(&header_only, 1024)
+            GzipProvider::default().decompress_with_limit(&MINIMAL_GZIP_HEADER, 1024)
         })
         .expect_err("header-only gzip member must be rejected");
         assert!(
@@ -1458,10 +1463,16 @@ mod tests {
             provider.decompress_with_limit(&compressed[..14], 1024)
         })
         .expect_err("truncated deflate stream must be rejected");
+        // Which check rejects the prefix depends on where the deflate encoder
+        // happened to place block boundaries: an incomplete block is caught by
+        // the stalled-stream handling, while a prefix that ends on a complete
+        // block is caught by the trailer-length check. Either way the
+        // truncated payload must be rejected.
+        let msg = err.to_string();
         assert!(
-            err.to_string()
-                .contains("truncated or invalid deflate stream"),
-            "unexpected error message: {err}"
+            msg.contains("truncated or invalid deflate stream")
+                || msg.contains("too short for trailer"),
+            "unexpected error message: {msg}"
         );
     }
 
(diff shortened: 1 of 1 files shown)


connectrpc/src/compression.rs
```diff
@@ -1379,6 +1390,115 @@ mod tests {
...
+    #[test]
+    fn test_gzip_decompress_truncated_deflate_stream() {
+        let provider = GzipProvider::default();
+        let data = b"hello world, this is a test of gzip compression";
+        let compressed = provider.compress(data).unwrap();
+
+        let err = run_with_timeout(TRUNCATION_TEST_TIMEOUT, move || {
+            // Keep the 10-byte header plus a prefix of the deflate stream,
+            // drop the rest (including the 8-byte trailer).
+            provider.decompress_with_limit(&compressed[..14], 1024)
```
rpb-ant, line 1463: This assertion is somewhat coupled to the backend's encoding: it relies on the level-1 output of this particular input having an incomplete block within `compressed[10..14]`. If a future flate2/zlib-rs bump changed the encoding so the first four deflate bytes happened to form a complete block, the failure would come from the trailer-length check instead (`"too short for trailer"`) and this assertion would fail even though the truncation handling is still correct.

Cheap way to harden it: accept either message here (`truncated or invalid deflate stream` *or* `too short for trailer`), since both mean the truncated payload was rejected — which is what this test is really about. The header-only test above already pins the `BufError` path specifically.

connectrpc/src/compression.rs
```diff
@@ -1379,6 +1390,115 @@ mod tests {
...
+    /// registry (the path the request/response handling code uses).
+    #[cfg(feature = "gzip")]
+    #[test]
+    fn test_gzip_registry_decompress_truncated() {
+        let registry = CompressionRegistry::new().register(GzipProvider::default());
+        let header_only = [0x1f, 0x8b, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xff];
+        let result = run_with_timeout(TRUNCATION_TEST_TIMEOUT, move || {
+            registry.decompress_with_limit("gzip", Bytes::copy_from_slice(&header_only), 1024)
+        });
+        assert!(result.is_err());
```
rpb-ant: The sibling tests assert the message; this one only asserts `is_err()`, so any error (e.g. a future header-validation rejection that never reaches the inflate loop) would still satisfy it. Since the point of this test is that the registry path surfaces the same truncation error, worth tightening to `expect_err(...)` + `contains("truncated or invalid deflate stream")` like the others.

connectrpc/src/compression.rs
```diff
@@ -1379,6 +1390,115 @@ mod tests {
...
+        );
+    }
+
+    /// A truncated gzip payload is also rejected when it arrives through the
+    /// registry (the path the request/response handling code uses).
+    #[cfg(feature = "gzip")]
+    #[test]
+    fn test_gzip_registry_decompress_truncated() {
+        let registry = CompressionRegistry::new().register(GzipProvider::default());
+        let header_only = [0x1f, 0x8b, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xff];
```
rpb-ant: Nit: same 10-byte literal as in `test_gzip_decompress_header_only` — could be a shared `const MINIMAL_GZIP_HEADER: [u8; 10]` next to `TRUNCATION_TEST_TIMEOUT`, with a one-liner naming the fields (id1, id2, CM=deflate, FLG=0, MTIME=0, XFL=0, OS=0xff). Optional.

rpb-ant (commented):
Took a careful look at this. The fix looks right and is at the right layer: the inflate loop's only exit was `StreamEnd`, so a deflate stream that ends without a final block spins forever on `BufError`, and only the provider can see that the decompressor is stuck. The "BufError ⇒ truncated" reasoning holds up — the loop always offers all remaining input and guarantees spare output capacity before each `decompress_vec` call, output-full progress comes back as `Ok`, and hard corruption surfaces as `Err(DecompressError)` through the existing `map_err` — so `BufError` here can only mean the input can't complete the stream. Making the match exhaustive over `flate2::Status` instead of using a wildcard is a nice touch, and the inline comment captures exactly the invariant the fix relies on.

A few test-level nits inline; none blocking. Two optional thoughts that didn't fit a line:

- The new error reuses the `"gzip decompression failed: ..."` prefix from the `map_err` just above. Something like `"gzip decompression stalled: truncated or invalid deflate stream"` would make the two failure modes easier to tell apart in logs. Take it or leave it.
- `internal` as the code for a truncated peer-supplied payload is consistent with every other malformed-gzip error in this function, so no change needed here — just noting that if there's ever an appetite for `invalid_argument` on malformed compressed payloads, that'd be a separate file-wide decision.

iainmcgin:
[claude code] Addressed in the latest commit: the stalled-stream error now reads "gzip decompression stalled: …" so it is distinguishable from the other failure modes in logs; the mid-stream truncation test accepts either rejection path (stalled-stream or trailer-length, since which one fires depends on where the encoder placed block boundaries); the registry-level test now asserts the error message; and the 10-byte header literal is a shared `MINIMAL_GZIP_HEADER` constant with the fields named. Agreed on `internal` vs `invalid_argument` being a separate file-wide decision — noted for a follow-up.

rpb-ant (approved):
Thanks — the reworded stalled-stream error, the either-path assertion in the mid-stream truncation test, the message check in the registry test, and the shared `MINIMAL_GZIP_HEADER` constant all look good. LGTM.

PR #133, opened 2026-05-27
client: bound Connect client-streaming response parsing at one message

iainmcgin:
## Summary

The Connect client-streaming response parser scanned every envelope in the collected response body: each data envelope was decompressed and accumulated, the "exactly one response message" rule was only enforced after the loop, and the END_STREAM envelope did not stop the scan. A response packed with many small compressed envelopes could therefore make the client perform per-envelope decompression and hold every decompressed payload before the count check ran, well beyond the single message the RPC shape allows.

The envelope scan now lives in a dedicated helper, `parse_connect_client_stream_envelopes`, which:

- rejects a second data envelope before its payload is decompressed,
- stops scanning at the END_STREAM envelope (the protocol-level end of the response), ignoring anything after it,
- returns the single message payload plus the END_STREAM metadata trailers, with the existing END_STREAM error handling unchanged.

Behavior for well-formed responses (one data envelope, then END_STREAM) is unchanged; the generic response function now just collects the body, calls the helper, and decodes the message.

## Tests

Four unit tests on the helper: second-message rejection happens before decompression (the second envelope is flagged compressed but contains garbage, so reaching the decompressor would produce a different error), scanning stops at END_STREAM with trailing bytes ignored and metadata trailers surfaced, END_STREAM errors still map to the right code/message, and a body with no data envelope is rejected. The client conformance suites in CI cover the well-formed paths.

### DIFF
--- a/connectrpc/src/client/mod.rs
+++ b/connectrpc/src/client/mod.rs
@@ -2992,8 +2992,43 @@ where
         .saturating_add(RESPONSE_BUFFER_TRAILER_SLACK);
     let body = collect_body_bounded(response.into_body(), body_limit).await?;
 
+    let (data, trailers) = parse_connect_client_stream_envelopes(
+        body,
+        &config.compression,
+        encoding.as_deref(),
+        max_msg_size,
+        &resp_headers,
+    )?;
+    let message = decode_response_view::<RespView>(data, config.codec_format)?;
+
+    Ok(UnaryResponse {
+        headers: resp_headers,
+        body: message,
+        trailers,
+    })
+}
+
+/// Scan a collected Connect client-streaming response body.
+///
+/// The body must contain exactly one data envelope followed by an END_STREAM
+/// envelope. Returns the (still encoded) message payload and any trailers
+/// carried in the END_STREAM metadata.
+///
+/// Scanning stops at the END_STREAM envelope — it is the protocol-level end
+/// of the response, so anything after it is ignored rather than decoded. A
+/// second data envelope is rejected before its payload is decompressed, so a
+/// response cannot make the client do per-envelope decompression work (or
+/// hold per-envelope decompressed buffers) beyond the single message the RPC
+/// shape allows.
+fn parse_connect_client_stream_envelopes(
+    body: Bytes,
+    compression: &crate::compression::CompressionRegistry,
+    encoding: Option<&str>,
+    max_msg_size: usize,
+    resp_headers: &http::HeaderMap,
+) -> Result<(Bytes, http::HeaderMap), ConnectError> {
     let mut buf = BytesMut::from(body.as_ref());
-    let mut data_envelopes: Vec<Bytes> = Vec::new();
+    let mut message: Option<Bytes> = None;
     let mut trailers = http::HeaderMap::new();
 
     while !buf.is_empty() {
@@ -3004,12 +3039,10 @@ where
 
         if envelope.is_end_stream() {
             let end_stream_data = if envelope.is_compressed() {
-                let enc = encoding.as_deref().ok_or_else(|| {
+                let enc = encoding.ok_or_else(|| {
                     ConnectError::internal("received compressed END_STREAM without encoding header")
                 })?;
-                config
-                    .compression
-                    .decompress_with_limit(enc, envelope.data, max_msg_size)?
+                compression.decompress_with_limit(enc, envelope.data, max_msg_size)?
             } else {
                 envelope.data
             };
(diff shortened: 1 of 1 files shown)


connectrpc/src/client/mod.rs
```diff
@@ -2992,8 +2992,43 @@ where
...
+    Ok(UnaryResponse {
+        headers: resp_headers,
+        body: message,
+        trailers,
+    })
+}
+
+/// Scan a collected Connect client-streaming response body.
+///
+/// The body must contain exactly one data envelope followed by an END_STREAM
```
rpb-ant: Tiny doc nit: as implemented, a body that ends after the data envelope with no END_STREAM at all is still accepted (message returned, empty trailers) — same leniency as the old loop. So "must contain exactly one data envelope followed by an END_STREAM envelope" overpromises slightly; something like "at most one data envelope; the END_STREAM envelope, if present, terminates the scan and supplies trailers" would match the behavior. (Or, if you'd rather enforce the stricter contract, that's a behavior change worth its own test.)

connectrpc/src/client/mod.rs
```diff
@@ -3039,57 +3072,55 @@ where
...
-                envelope.data
-            };
 
-            if data.len() > max_msg_size {
-                return Err(ConnectError::new(
-                    ErrorCode::ResourceExhausted,
-                    format!("message size {} exceeds limit {}", data.len(), max_msg_size),
-                ));
-            }
+            // END_STREAM is the end of the logical response stream; stop
```
rpb-ant, line 3082: Optional: the request-side analogue (#130) drains post-END_STREAM bytes and logs a `tracing::warn!` once when there's trailing data; here trailing bytes are dropped silently. A one-shot `tracing::debug!` when `buf` is non-empty after this break would keep the leniency while making a misbehaving server visible. Fine to skip if you'd rather keep the parser quiet — the silent ignore is also consistent with how `StreamReader` treats END_STREAM on the streaming path.

connectrpc/src/client/mod.rs
```diff
@@ -4430,4 +4461,132 @@ mod tests {
...
+    fn client_stream_response_rejects_second_message_before_decompression() {
+        let registry = crate::compression::CompressionRegistry::default();
+
+        let mut body = Envelope::data(Bytes::from_static(b"first"))
+            .encode()
+            .to_vec();
+        let mut bogus = Envelope::data(Bytes::from_static(b"not gzip data"))
+            .encode()
+            .to_vec();
+        bogus[0] |= 0x01; // compressed flag
```
rpb-ant: Nit: `Envelope::compressed(...)` already exists, so this could be

```rust
let bogus = Envelope::compressed(Bytes::from_static(b"not gzip data")).encode();
```

which drops the hand-rolled flag flip (and the `0x01` magic number) and reads more directly as "a compressed envelope whose payload won't decompress". The property the test relies on — reaching the decompressor would produce a different error — is unchanged.

rpb-ant (commented):
This looks good. The body is already collected under a hard cap, but the parser is the only layer that knows the RPC shape, and the old loop decompressed and held every data envelope before the count check ran — so a ~4 MiB body packed with small compressed envelopes could drive a lot of decompression work and held memory for a response that's only allowed one message. Failing on the second data envelope before touching its payload and treating END_STREAM as terminal is the right call, and tracing the new helper against the old loop, the well-formed path looks preserved exactly: error codes and messages, header/trailer propagation, the `Unimplemented → Internal` rewrite for data-envelope decompression failures, and the END_STREAM error mapping all match. Extracting the helper also finally makes this unit-testable without transport mocks — the "second envelope flagged compressed but containing garbage" fixture is a nice way to prove decompression is never reached.

Non-blocking notes beyond the inline ones:

- `parse_grpc_unary_response` still has the same shape of issue (per-envelope decompression with the count check after the loop). It's milder there — it overwrites rather than accumulates — but worth a follow-up so the gRPC/gRPC-Web path gets the same fail-fast treatment.
- The `unimplemented` code for zero/multiple messages predates this PR, but the new tests now pin it. connect-go uses `unknown` for the same condition; conformance passes either way, so no change requested — just worth being deliberate that we're standardizing on `unimplemented`.
- Optional extra unit tests now that the helper is isolated: a compressed END_STREAM envelope (flags `0x03`), one happy-path case with a compressed data envelope, and END_STREAM-before-data — which is now rejected as "no data messages", whereas the old loop would have accepted a data envelope appearing after END_STREAM, so a test would pin that the new behavior is intended.

iainmcgin:
[claude code] Addressed in the latest commit: the helper docs now say a body without an END_STREAM envelope is still accepted (matching the previous behaviour); trailing response data after END_STREAM is logged at debug level before being ignored; the second-message test uses `Envelope::compressed(...)` instead of the hand-rolled flag flip; and there are new tests for compressed data/END_STREAM envelopes and for a data envelope that only appears after END_STREAM (now rejected as "no data messages", pinning that behaviour change as intended).

On the broader notes: agreed that `parse_grpc_unary_response` deserves the same fail-fast treatment — queued as a follow-up PR rather than widening this one; and standardizing on `unimplemented` for the zero/multiple-message cases is deliberate for now (the new tests pin it, so changing to `unknown` later would be an explicit decision).

rpb-ant (approved):
Thanks — the doc now matches the actual leniency around a missing END_STREAM, the debug log makes trailing data visible without changing behavior, the `Envelope::compressed` fixture reads better, and the two new tests (compressed data/END_STREAM envelopes, data-after-END_STREAM rejected) pin exactly the cases that were open questions. Sounds good on the `parse_grpc_unary_response` follow-up. LGTM.

Issue #129, opened 2026-05-21
Are there any plans for implementing gRPC server reflection?


torkelrogstad:
Thanks for a great implementation of Connect in Rust. Eagerly migrating over a gRPC service right now, and I'm very happy so far. The one big issue is that reflection is not supported. Is this something you have concrete plans for? I'd be happy to make an attempt at it, like I've done with #128 . Reflection is a lot more complex than health checks, though. Figured I'd first ask if you have any thoughts for how it should be done, directionally.

iainmcgin:
buffa 0.7.0 has a full reflection implementation for messages - once #143 is merged (picked up 0.7.0 and restructures some of the API surface to be compatible), we should be in a better position to implement support for some reflection based features at the Connect / gRPC level. I think the main use case is for generic interceptors that work for any RPC, interpreting message or field annotations. Do you have other use cases I should be aware of and try to support?

torkelrogstad:
I believe we're talking about different kinds of reflection. I see now that my original message was imprecise. I'm talking about an equivalent to [`connectrpc/grpcreflect-go`](https://github.com/connectrpc/grpcreflect-go).

iainmcgin:
Ah - thanks for the clarification. Yes, I think it's reasonable for us to add a standard crate for this similar to your connectrpc-health create, and #141 also introduces some of the machinery we'd need to expose the descriptor sets required for reflective tooling. I'll spike this with Claude just now and see if it looks straightforward.

iainmcgin:
[claude code] Implemented by #157 (merged as 218ac59): the connectrpc-reflection crate provides grpc.reflection.v1 + v1alpha, self-describing, with both FileDescriptorSet-bytes and DescriptorPool sources. See the multiservice example and examples/multiservice/reflection-demo.sh for an end-to-end buf curl walkthrough. Closing manually since auto-close is not enabled.

PR #141, opened 2026-05-27
connectrpc-build: add Config::emit_descriptor_set for gRPC server reflection

christopherwxyz:
# connectrpc-build: add `Config::emit_descriptor_set` for gRPC Server Reflection

## Motivation

`Config::compile()` already parses the `.proto` inputs into a full `FileDescriptorSet` — `run_protoc` is invoked with `--include_imports`, `run_buf` with `--as-file-descriptor-set`, and the precompiled path is read as-is, so the `descriptor_bytes` it decodes for codegen always carry the **complete transitive import closure**.

Today that set can be *read in* (`Config::descriptor_set(path)`) but never *written out*. A build script that needs those bytes — to serve `grpc.reflection.v1.ServerReflection` to clients like `grpcurl` via `include_bytes!(concat!(env!("OUT_DIR"), "…"))` — has to invoke `protoc` a **second** time with `--descriptor_set_out --include_imports`, duplicating work connectrpc-build already did and re-requiring a `protoc` install even when generation used `buf` or a precompiled set.

This adds the symmetric `Config::emit_descriptor_set(name)`: write the descriptor set connectrpc-build already computed to `<out_dir>/<name>`.

This is the connectrpc-build counterpart of the [buffa#125 follow-up the maintainers invited](https://github.com/anthropics/buffa/issues/125#issuecomment-4514370502) — buffa-build went with the embedded `buffa-reflect` pool for in-process reflection, explicitly leaving open the "standalone `.bin` for tooling outside the Rust build" case. Emitting at the connectrpc-build layer is the natural home for that: the bytes are already in hand, with `--include_imports` guaranteed, for every descriptor source.

## API

```rust
/// Also write the parsed `FileDescriptorSet` (the descriptors used for
/// codegen) to `<out_dir>/<name>` as wire-format bytes.
///
/// The set carries the full transitive import closure for every descriptor
/// source (`protoc --include_imports`, `buf --as-file-descriptor-set`, or a
/// precompiled set), so it is ready to back `grpc.reflection.v1.ServerReflection`.
/// The inverse of [`Config::descriptor_set`], which *reads* a precompiled set.
#[must_use]
pub fn emit_descriptor_set(mut self, name: impl Into<String>) -> Self
```

```rust
// build.rs
connectrpc_build::Config::new()
    .files(&["proto/svc.proto"])
    .includes(&["proto/"])
    .emit_descriptor_set("svc_descriptor.bin")
    .compile()?;

// src/lib.rs
pub const FILE_DESCRIPTOR_SET: &[u8] =
    include_bytes!(concat!(env!("OUT_DIR"), "/svc_descriptor.bin"));
```

## Implementation (connectrpc-build/src/lib.rs)

Purely additive — no change to existing behavior; the field defaults to `None`.

1. `Config` struct — new field:
```rust
     include_file: Option<String>,
+    emit_descriptor_set: Option<String>,
     emit_rerun_directives: bool,
```
2. `Config::new()` — default:
```rust
             include_file: None,
+            emit_descriptor_set: None,
             emit_rerun_directives: true,
```
3. Builder method (next to `descriptor_set`):
```rust
+    #[must_use]
+    pub fn emit_descriptor_set(mut self, name: impl Into<String>) -> Self {
+        self.emit_descriptor_set = Some(name.into());
+        self
+    }
```
4. In `compile()`, right after `create_dir_all(&out_dir)` (with `descriptor_bytes` already in scope from the descriptor-source match):
```rust
+        // Emit the parsed descriptor set for reflection, if requested.
+        if let Some(name) = &self.emit_descriptor_set {
+            write_if_changed(&out_dir.join(name), &descriptor_bytes)?;
+        }
```

It reuses the existing `write_if_changed` (so it won't churn mtimes / trigger needless rebuilds), and works uniformly for the `Protoc`, `Buf`, and `Precompiled` sources because each already produces an import-complete `descriptor_bytes`.

## Tests (included)

Two tests added to the existing `mod tests`, mirroring `compile_precompiled_descriptor_set`:
- `config_emit_descriptor_set_toggle` — builder-state check.
- `emit_descriptor_set_writes_reflection_bin` — end-to-end against the `echo.fds.bin` fixture (hermetic, Precompiled source, no protoc): asserts the file is written, `FileDescriptorSet::decode_from_slice` succeeds with a non-empty file list, and the bytes round-trip the source set exactly.

Verified locally: `cargo build/clippy -p connectrpc-build` clean; `cargo test -p connectrpc-build` → **20 passed, 0 failed**.

## Notes
- Additive and backward-compatible; no existing API changes.
- Complements `buffa-reflect`'s in-process pool — this is the wire-bytes path for serving reflection v1 to external clients.

Follow-up to anthropics/buffa#125.

### DIFF
--- a/connectrpc-build/src/lib.rs
+++ b/connectrpc-build/src/lib.rs
@@ -62,6 +62,7 @@ pub struct Config {
     out_dir: Option<PathBuf>,
     descriptor_source: DescriptorSource,
     include_file: Option<String>,
+    emit_descriptor_set: Option<String>,
     emit_rerun_directives: bool,
     options: Options,
 }
@@ -75,6 +76,7 @@ impl Config {
             out_dir: None,
             descriptor_source: DescriptorSource::default(),
             include_file: None,
+            emit_descriptor_set: None,
             emit_rerun_directives: true,
             options: Options::default(),
         }
@@ -220,6 +222,35 @@ impl Config {
         self
     }
 
+    /// Also write the parsed `FileDescriptorSet` (the descriptors used for
+    /// codegen) to `<out_dir>/<name>` as wire-format bytes.
+    ///
+    /// The set carries the full transitive import closure for every descriptor
+    /// source (`protoc --include_imports`, `buf --as-file-descriptor-set`, or a
+    /// precompiled set), so it is ready to back `grpc.reflection.v1.ServerReflection`
+    /// for clients such as `grpcurl`. Pair it with `include_bytes!`:
+    ///
+    /// ```ignore
+    /// // build.rs
+    /// connectrpc_build::Config::new()
+    ///     .files(&["proto/svc.proto"])
+    ///     .includes(&["proto/"])
+    ///     .emit_descriptor_set("svc_descriptor.bin")
+    ///     .compile()?;
+    /// // src/lib.rs
+    /// pub const FILE_DESCRIPTOR_SET: &[u8] =
+    ///     include_bytes!(concat!(env!("OUT_DIR"), "/svc_descriptor.bin"));
+    /// ```
+    ///
+    /// The inverse of [`Config::descriptor_set`], which *reads* a precompiled
+    /// set; this *writes* the one connectrpc-build already computed, so build
+    /// scripts no longer need a second `protoc --descriptor_set_out` pass.
+    #[must_use]
+    pub fn emit_descriptor_set(mut self, name: impl Into<String>) -> Self {
+        self.emit_descriptor_set = Some(name.into());
+        self
+    }
+
     /// Emit an `include!`-based module tree file alongside the per-file
     /// `.rs` outputs.
     ///
@@ -304,6 +335,13 @@ impl Config {
         //    stitchers.
         std::fs::create_dir_all(&out_dir)
             .with_context(|| format!("failed to create out_dir '{}'", out_dir.display()))?;
+
+        // Emit the parsed descriptor set for gRPC server reflection, if requested.
+        // `descriptor_bytes` already carries the full import closure for every
+        // descriptor source, so the written set is reflection-ready as-is.
+        if let Some(name) = &self.emit_descriptor_set {
+            write_if_changed(&out_dir.join(name), &descriptor_bytes)?;
+        }
 
         let mut entries: Vec<(String, String)> = Vec::new();
         for file in &generated {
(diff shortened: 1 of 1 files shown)


christopherwxyz:

christopherwxyz:
recheck

iainmcgin (changes requested):
[claude code] Thanks — this is a well-built building block, and well-timed: we've just committed (in #129) to spiking a `connectrpc-reflection` crate, and this is a prerequisite for it. Two review agents went over the change; the core design is right, in particular writing the raw `descriptor_bytes` rather than a re-encode (preserves unknown fields; byte-identical on the precompiled path), placing the write after codegen (no stale file on failure), and reusing `write_if_changed` (no rebuild churn).

A few findings need addressing before merge. Normally I'd push these to your branch directly, but GitHub doesn't allow maintainer edits on organization-owned forks, so I've staged everything as a single commit on `suggestions/pr-141-hardening` in this repo — please cherry-pick it onto your branch (it's based on your head after a merge of current `main`):

```sh
git remote add upstream https://github.com/anthropics/connect-rust.git # if needed
git fetch upstream main suggestions/pr-141-hardening
git merge upstream/main          # picks up the changelog sections added since you branched
git cherry-pick aafd9ad          # the hardening commit
git push
```

What the commit contains:

1. **`name` validation** (Medium): the doc promises `<out_dir>/<name>`, but `Path::join` silently escapes `out_dir` for absolute paths or names containing separators. Names with separators are now rejected with a clear error, the write gets path context (`with_context`), and `compile()`'s `# Errors` list mentions the new failure mode.
2. **Import-closure test** (Medium): both existing fixtures are single-file with no imports, so the tests never exercised the feature's headline property. Added `imports.fds.bin` (`uses_dep.proto` importing `dep.proto`, built with `protoc --include_imports`) and a test asserting both files appear in the emitted set; the existing test now asserts file names instead of non-empty.
3. **Changelog entry** (required for user-facing changes) and a README update — the "Not yet implemented: gRPC server reflection" line now points at this method and #129.

One advisory item we noted but did not change: repeat calls (`emit_descriptor_set("a").emit_descriptor_set("b")`) silently keep only the last name. That matches `include_file` / `out_dir` semantics, so it's fine — and if the reflection crate ever wants multiple sets, that's an additive method, not a break.

Once the cherry-pick is pushed this is approve-and-merge ready from our side — though heads-up on timing: #143 (a large breaking change) lands first for 0.7.0, after which this will need a trivial changelog re-merge, which we're happy to handle.

iainmcgin:
[claude code] Heads-up: GitHub doesn't permit maintainer pushes to organization-owned forks, so we've carried this forward as #156 — your original commit plus the hardening from the review above, with you credited as co-author on the squash commit. This PR can be closed in favor of #156. Thanks again for the contribution, and watch #129 if you're interested in the reflection crate this unblocks.

PR #142, opened 2026-05-27
examples/bazel: regenerate Cargo.lock against connectrpc 0.6.1

iainmcgin:
Updates the bazel example's lockfile to the published connectrpc 0.6.1 (the dependency requirement was already bumped in #134). buffa stays at 0.6.0. Same post-release follow-up as #124 was for 0.6.0.


iainmcgin:
Closing - will rework this after the 0.7.0 release

PR #144, opened 2026-05-29
compression: report malformed payloads as invalid_argument

### LINKED ISSUE #139: compression: error code for malformed compressed payloads
All malformed-compressed-input errors in the gzip/zstd providers (bad header, truncated stream, CRC mismatch, size mismatch, etc.) are currently reported as `internal`. For a request whose compressed payload is malformed, the fault lies with the sender, and `invalid_argument` (or `data_loss`) would describe it better; `internal` also skews server-error metrics for what is structurally a client error.

This is a deliberate, file-wide decision rather than a one-off change: it affects every error site in `connectrpc/src/compression.rs` and is observable to clients, so it should be made consistently and noted in the changelog. Conformance does not constrain the choice either way.

Yong-yuan-X:
## Summary

- report malformed gzip and zstd compressed payloads as `invalid_argument` instead of `internal`
- apply the error-code change consistently across gzip/zstd bad header, truncated stream, CRC mismatch, size mismatch, and decoder failure paths
- add regression coverage for malformed gzip/zstd payload error codes
- document the observable error-code change in the changelog

Closes #139.

## Testing

- `cargo test -p connectrpc compression`


iainmcgin:
[claude code] Thanks for the contribution — the server-side change is exactly what #139 asked for, and the regression tests asserting the actual error codes (not just `is_err()`) are appreciated.

I pushed two commits to the branch as maintainer edits:

1. **`client: remap malformed response payloads to data_loss`** — the compression providers are shared between the server (decompressing requests) and the client (decompressing responses). With this PR as it stood, a corrupt *server response* would surface to client callers as `invalid_argument`, which misleadingly tells the caller their request was bad. The client already remapped `unimplemented` → `internal` on the response path for the same reason, so I folded that into a shared `map_response_decompression_error` helper that also remaps `invalid_argument` → `data_loss`, and applied it at all five response-decompression sites (including the gRPC unary and Connect END_STREAM sites that previously propagated provider codes unmapped).

   Note that this deliberately diverges from connect-go, which reports `invalid_argument` in both directions — we feel `data_loss` ("unrecoverable data loss or corruption") is more descriptive of what actually happened to a corrupt response, without implying the request was at fault.

2. **A merge of `main`** to pick up #147 and resolve the overlapping changelog sections. The changelog entry now also covers the 500 → 400 monitoring implication for server operators and the client-side remap.

iainmcgin (approved):

PR #146, opened 2026-05-30
client: error on truncated Connect streaming responses

### LINKED ISSUE #140: client: surface truncated Connect streaming responses as an error
For Connect-protocol streaming responses, `ServerStream` treats the body ending without an END_STREAM envelope as a clean end of stream: `message()` returns `Ok(None)` and `error()` is `None`. A response that is cut off mid-stream (connection drop, misbehaving intermediary) is therefore indistinguishable from a complete one. The gRPC and gRPC-Web paths already surface missing trailers as an error, and connect-go reports a missing END_STREAM as an error as well.

Proposed work: when the body ends while the protocol is Connect and no END_STREAM envelope has been seen, surface an error (e.g. `unavailable` or `internal`, matching connect-go's choice) instead of a clean end, with a regression test for the truncated-response case.

Yong-yuan-X:
## Summary

- return an error when a Connect streaming response body ends before an END_STREAM envelope is received
- add a regression test for a response that yields a data envelope and then EOFs without END_STREAM
- document the client-visible behavior change in the changelog

Closes #140.

## Testing

- `cargo check -p connectrpc --all-features --all-targets`
- `cargo clippy -p connectrpc --all-features --all-targets -- -D warnings`
- `cargo test -p connectrpc connect_server_stream_truncated_after_data_errors`
- `cargo test -p connectrpc --no-default-features`
- `RUSTDOCFLAGS='-Dwarnings' cargo doc -p connectrpc --all-features --no-deps`


iainmcgin (approved):
[claude code] Reviewed by both review agents: the truncation detection correctly distinguishes a body that ends after END_STREAM from one cut off without it, is gated to the Connect protocol only, and the unavailable code matches connect-go. I pushed a follow-up to the branch with the empty-body regression test, an # Errors note on message(), an expanded changelog entry flagging the client-visible behavior change, and a merge of main. Thanks for the contribution!

PR #147, opened 2026-05-30
client: tighten gRPC and gRPC-Web unary response parsing

### LINKED ISSUE #138: client: tighten gRPC and gRPC-Web unary response parsing
Two follow-ups from the recent client-streaming response parser work (#133), where the same shape exists in milder form:

- `parse_grpc_unary_response` decompresses every data envelope it encounters and only enforces the single-message expectation afterwards (later envelopes overwrite earlier ones rather than accumulating). Rejecting a second data envelope before decompressing it would match the Connect client-streaming behaviour.
- The gRPC-Web response collection loop buffers frames until body EOF or the size cap and only then looks for the trailers frame (flag 0x80). Stopping at the trailers frame would avoid buffering and then failing on data a server sends after it, and would let well-formed responses complete even if the server keeps writing.

Both are bounded today by the existing response-buffer caps; this is about failing fast and doing no more decompression work than the RPC shape allows.

jeet-dekivadia:
## Summary
- stop collecting gRPC-Web unary body frames once a complete trailers frame is buffered
- reject a second unary response message before attempting decompression
- add focused regressions for both parser paths

## Why
Unary and client-streaming responses have a single-message shape. Failing fast avoids unnecessary decompression work and lets well-formed gRPC-Web responses finish without waiting for body EOF after trailers arrive.

## Validation
- `git diff --check`
- `cargo +nightly fmt --all -- --check`

Fixes #138

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,20 @@ with the [Rust 0.x convention](https://doc.rust-lang.org/cargo/reference/semver.
 breaking changes increment the minor version (0.2 → 0.3), additive changes
 increment the patch version.
 
+## [Unreleased]
+
+### Fixed
+
+- **The gRPC and gRPC-Web unary response parsers enforce the
+  single-message rule before decompressing, and gRPC-Web parsing stops at
+  the trailers frame** ([#147]). A second data envelope is rejected before
+  its payload is touched (matching the Connect client-streaming parser
+  from [#133]), and a gRPC-Web response now completes as soon as a
+  complete trailers frame is buffered instead of reading the body to EOF,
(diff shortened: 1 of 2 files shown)


jeet-dekivadia:

iainmcgin (approved):
[claude code] Reviewed by both code-review agents: the second-envelope rejection correctly fires before decompression with the same error code as the Connect path, and the trailers-frame scanner handles partial frames, compressed trailers, and post-trailer data correctly. I pushed a small follow-up commit to the branch adding the changelog entry and making the multiple-message error shape-neutral (the parser also serves client-streaming responses). Thanks for the contribution!

PR #148, opened 2026-05-30
server: include body receipt in request deadlines

### LINKED ISSUE #136: server: include request body receipt in the request deadline
For unary and server-streaming RPCs the request body is collected before the deadline wrapper is applied (`collect_body_limited` runs ahead of the `tokio::time::timeout` call in `connectrpc/src/service.rs`), so a configured `DeadlinePolicy` bounds only handler execution, not the time spent receiving the request. A request that sends headers and then delivers its body very slowly holds the request future for as long as the connection stays open, regardless of policy. Client-streaming and bidi RPCs are not affected because the body is read inside the handler.

Proposed work: start the deadline clock before body collection (or wrap the body read in the same timeout), so that `with_default_timeout` / `with_max` bound the whole request. Worth doing together with the connection-level timeout work, since with the default no-op policy there is currently no time bound at all.

jeet-dekivadia:
## Summary
- anchor buffered request deadlines before body collection so one absolute budget covers receipt and handler execution
- apply the same bounded collection to error-path drains while leaving client-streaming and bidi request flow unchanged
- document the request-receipt guarantee and add paused-time regressions for stalled unary and server-streaming bodies

Fixes #136

## Validation
- `git diff --check`
- `cargo +nightly fmt --all -- --check`


iainmcgin (dismissed):
[claude code] Reviewed by both review agents: the deadline is anchored once as an absolute instant and shared across body receipt, handler execution, and the response stream, the deadline visible through RequestContext now matches what is enforced, and client/bidi paths are untouched. I pushed a follow-up to the branch with the changelog entry, rustdoc updates for with_max / with_default_timeout, the stale guide line fix, and a merge of main. Thanks for the contribution!

iainmcgin (approved):
[claude code] Re-approving after the main merge (changelog conflict with #150 resolved; both entries kept).

PR #150, opened 2026-06-05
service: surface streaming request body read errors to handlers

EffortlessSteven:
## What this does

A request body transport failure (`Body::poll_frame` returning `Err`) mid-decode made `spawn_body_reader` stop, ending the handler's request stream with a clean EOF. A truncated client-streaming/bidi upload was then indistinguishable from a complete stream, so a handler aggregating messages treated partial input as a finished stream.

**Fix:** while still decoding, forward the failure to the handler stream as `ConnectError::internal` (`BodyReader::on_body_error`), the same error the unary body-read path already returns. Once only draining trailing bytes (after END_STREAM, a decode error, or a dropped receiver), it stays diagnostic-only.

## Verification

Red on `main`, green after: unpatched, the decoding-path test sees `None` where an `Err` belongs.

| Check | Result |
| --- | --- |
| body error while decoding | unpatched `None` (clean EOF), patched `Err(Internal)` after the message |
| body error after END_STREAM | clean end, no spurious `Err` |
| `cargo +nightly fmt --all --check` | clean |
| `cargo clippy -p connectrpc --all-features --all-targets -- -D warnings` | clean |
| `cargo test -p connectrpc --all-features` | 410 passed |

## Review map

- `connectrpc/src/service.rs`: `on_body_error` surfaces the failure while `Decoding`, suppresses it while `Draining`; read loop calls it instead of dropping the cause.
- `connectrpc/src/service.rs` (tests): pin the surfaced and suppressed paths.


EffortlessSteven:

iainmcgin (approved):
[claude code] Reviewed by both review agents: the decode-vs-drain gating is correct (no decode-phase error can be swallowed, no post-END_STREAM error spuriously surfaces), the internal code matches the unary body-read path exactly, and the bounded-drain behavior from #130 is preserved. I pushed a follow-up to the branch with the changelog entry, None-vs-Err documentation on ServiceStream and in the guide, a comment recording the error-code rationale, and a merge of main. Thanks for the contribution!

PR #157, opened 2026-06-09
Add connectrpc-reflection crate (gRPC server reflection v1 + v1alpha)

### LINKED ISSUE #129: Are there any plans for implementing gRPC server reflection?

iainmcgin:
## Summary

Adds the `connectrpc-reflection` crate: gRPC server reflection (`grpc.reflection.v1.ServerReflection` and its `v1alpha` predecessor) for connectrpc servers, closing the gap tracked in #129. Reflection-aware tooling — `grpcurl`, `buf curl`, Postman, `grpcui` — can discover and call services over gRPC, gRPC-Web, and the Connect protocol alike.

## Design

A `Reflector` answers the five reflection queries by delegating name resolution (symbols, extensions, services) to `buffa_descriptor::DescriptorPool`, whose symbol index was built for exactly this protocol. Two descriptor sources are supported:

- **`FileDescriptorSet` bytes** (`Reflector::from_descriptor_set_bytes`), e.g. the output of `connectrpc_build::Config::emit_descriptor_set` from #156. Responses carry the compiler's **original** per-file descriptor bytes, sliced out of the input by a small top-level wire walk — never a re-encode — so payloads from newer compilers survive byte-for-byte (pinned by a test). No reflection codegen is required for this path.
- **An existing `DescriptorPool`** (`Reflector::from_descriptor_pool`), e.g. the `descriptor_pool()` accessor that reflection-enabled buffa codegen emits, sparing the build script a separate emit step. Responses re-encode the pool's parsed files (unknown fields preserved).

`ReflectionService` bridges the index to the generated bidi-streaming traits for both protocol versions — the messages are wire-identical, so one macro body instantiates both impls — and `install(router, reflector)` mounts both. `Reflector::with_services` curates the advertised service list (the Go `grpcreflect` `Namer` equivalent). Lookup misses answer in-band with `NOT_FOUND` while malformed requests terminate the stream with `invalid_argument`, matching grpc-go.

The service is **self-describing**: every query falls back to the crate's own descriptors, so `ServerReflectionInfo` itself is resolvable and `ListServices` includes the reflection services. This matches grpc-go (where the reflection proto is always in the registry) and is load-bearing for schema-free clients — `buf curl` cannot call the reflection RPC without it. tonic requires manually registering its `FILE_DESCRIPTOR_SET` for the same effect; here it is the default, and `with_services` can de-list the reflection names if unwanted.

The crate mirrors `connectrpc-health`'s structure — checked-in `buf generate` output (`task connectrpc-reflection:generate`, wired into `generate:all`), a `client` feature gating the generated `ServerReflectionClient`, and a `wire` module re-exporting the message types downstream clients need — with one difference: instead of vendoring the external proto sources, both code generation and the checked-in descriptor set pull `grpc/reflection/{v1,v1alpha}` from the public Buf Schema Registry (`buf.build/grpc/grpc`), the same pattern the conformance crate uses; the registry is only needed at regeneration time. The crate also exposes `FILE_DESCRIPTOR_SET` (its own protos as an encoded set) so servers can make the reflection service describe itself — schema-free clients need that to invoke `ServerReflectionInfo` directly, and it is the equivalent of registering tonic-reflection's constant of the same name.

## Example integration

The multiservice example now mounts reflection, with **both descriptor sources** selectable via `REFLECTION_SOURCE`:

- `fds` (default) — a descriptor set produced by `buf build` in `task example:multiservice:generate` and checked in under `descriptor/` (the examples deliberately have no `build.rs`; a build-script project would use `emit_descriptor_set` instead — the server comments spell out both paths). The set lives outside `src/generated/` so the CI stale-check job is not sensitive to byte-level drift between buf releases.
- `pool` — the `descriptor_pool()` accessor emitted by buffa codegen with `reflect_mode=bridge` (now enabled in the example's `buf.gen.yaml`); no descriptor file involved at all. This required `strategy: all` on the buffa plugin: under buf's default per-directory strategy each generated package embeds only its own package's closure, so a single package's pool would not cover the other services — the buf.gen.yaml comment and `from_descriptor_pool`'s docs both record the trap.

`examples/multiservice/reflection-demo.sh` (also `task example:multiservice:reflection-demo`) walks through the protocol end to end with `buf curl` and zero local proto files: ListServices, fetching the file that declares `GreetService`, calling `Greet` and `Add` with schemas resolved via reflection, the same call over the legacy `v1alpha` reflect protocol, and an unknown-symbol query answered in-band with `NOT_FOUND`.

## Validation

- 19 unit/integration tests, including full bidi round trips through the generated client against a live server, symbol-scoping edge cases (enum values resolve in the enum's parent scope), merge dedup, the byte-exactness guarantee, and the self-describing descriptor set.
- `examples/multiservice/reflection-demo.sh` passes end to end; the example's existing `test.sh` (run by CI) still passes with reflection mounted.
- Verified end to end with `buf curl` as an independent client: schema resolution through the reflection service itself, calls over gRPC (h2c), gRPC-Web, and Connect, and the `grpc-v1alpha` fallback reflection protocol.
- `cargo clippy --all-targets --all-features -- -D warnings`, nightly `fmt`, and `cargo doc` all clean; workspace check passes.

Closes #129.


torkelrogstad:
Thanks for implementing this @iainmcgin! I've tested this branch in a migration of an existing gRPC service, and it works great. 

If you're interested in some input from me: I think checking in a binary file descriptor set is an antipattern. "Random" binary blobs are suspicious and not something that people should be encouraged to throw around. You could argue that the file-descriptor-in-rust-code pattern is basically the same thing, but at least it is possible to open the file in a normal editor and understand what it is.

iainmcgin:
> If you're interested in some input from me: I think checking in a binary file descriptor set is an antipattern. 

Is this in reference to the `reflection.fds.bin` file that is committed for the minimal test server? I generally agree and prefer not to check in binary / inscrutable files (also generated code, normally, but it's useful to retain generated code here so we can clearly see how codegen changes things). I think this was just something Claude reached for as an easy option to build upon the emit_descriptor_set support from #156, and that checking in the file also makes this easier for others to run (with the recreation steps in a comment if they are suspicious or want to know how it is created).

In general I think I prefer the approach here of using a DescriptorPool from buffa's reflection support, but the FDS approach also seems common so it seemed worth it to support both.

rpb-ant (approved):

PR #158, opened 2026-06-10
release: 0.7.0

iainmcgin:
## Summary

Prepares the 0.7.0 release: lockstep version bumps for five crates, changelog restructuring and finalization, a documentation accuracy pass across README / guide / rustdoc, and one small behavior fix surfaced by the pre-release audit.

### Release mechanics

- All five crates bump to 0.7.0: `connectrpc`, `connectrpc-codegen`, `connectrpc-build`, plus `connectrpc-health` (#128) and `connectrpc-reflection` (#157), which both join the lockstep scheme with this as their first published version. Inter-crate version requirements move to `0.7`.
- `publish-crates.yml` gains publish steps for `connectrpc-health` and `connectrpc-reflection` — the workflow previously published only three crates, so a tag push would have silently skipped both new crates.
- Both new crates gain docs.rs metadata (`all-features` + `--cfg docsrs`) and feature badges on their `client`-gated re-exports, matching the `connectrpc` pattern.

### Changelog

The `[Unreleased]` section accumulated two stacked fragments (the #143 conversion plus the #159/#160 entries appended above it). Restructured into one coherent `[0.7.0] - 2026-06-10` entry:

- Single intro and Breaking section covering both breaking fronts: the server-side request rework (`ServiceRequest`/`StreamMessage` on buffa 0.7.0) and the client streaming terminal-error contract, with a before/after handler-signature snippet and PR references on the streaming entries.
- Previously missing entries added: `connectrpc-health` (#128) and the inter-message timer fix (#127, including the corrected first-poll arming semantics).
- The #159 gRPC entry now cross-references #140 instead of implying the Connect `unavailable` behavior predates this release.

### Documentation accuracy pass

- Guide: new **Server reflection** section (dependency snippet, both descriptor sources, schema-exposure callout, demo pointers), fifth crate-table row, streaming-client section now states the #159 `Err`-on-terminal-error contract, bidi snippet fixed to unwrap the `Option` before `.reborrow()`, inter-message timeout wording updated for #127, `ctx.deadline` field access replaced with the real accessors, `ConnectError` sketch corrected, stale `0.5`/`0.6` version snippets bumped.
- Rustdoc: `connectrpc-reflection`'s build.rs quick-start used the prost-style `compile(&files, &includes)` signature that doesn't exist — fixed to the real `Config` API; `with_inter_message_timeout` now documents when the timer arms; `connectrpc` crate-doc module index and Protocol Support section refreshed; duplicated `ServerStream` summary line removed.

### Behavior fix

The default `CompressionProvider::decompress_with_limit` implementation mapped read failures to `internal`, while #144 moved the built-in providers to `invalid_argument` — a custom provider relying on the default would silently keep the old 5xx behavior the changelog says was fixed. The default now maps read failures to `invalid_argument`, with a documented convention and a regression test.

## Verification

- `task lint` (clippy all targets `-D warnings` + nightly fmt) clean
- `task test` passes
- `task doc` (broken intra-doc links denied) clean

## After merge

Tag `v0.7.0` on the merge commit to trigger crates.io publish (now five crates) and the binary release. Post-release follow-up: regenerate `examples/bazel/Cargo.lock` against the published 0.7.0 crates (its pins cannot resolve before publish). Optional: pick up buffa 0.7.1 in lockfiles when it ships — the `0.7` floor needs no change.

### DIFF
--- a/.github/workflows/publish-crates.yml
+++ b/.github/workflows/publish-crates.yml
@@ -98,3 +98,9 @@ jobs:
 
       - name: Publish connectrpc-build
         run: /tmp/publish.sh connectrpc-build $DRY_RUN_FLAG
+
+      - name: Publish connectrpc-health
+        run: /tmp/publish.sh connectrpc-health $DRY_RUN_FLAG
+
+      - name: Publish connectrpc-reflection
+        run: /tmp/publish.sh connectrpc-reflection $DRY_RUN_FLAG
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,63 +8,22 @@ with the [Rust 0.x convention](https://doc.rust-lang.org/cargo/reference/semver.
 breaking changes increment the minor version (0.2 → 0.3), additive changes
 increment the patch version.
 
-## [Unreleased]
+## [0.7.0] - 2026-06-10
 
-### Breaking
+A breaking release that reworks both ends of the message surface:
+handlers take borrowed request views built on buffa 0.7.0, and client
+streams surface terminal RPC errors from `message()` instead of hiding
+them behind `Ok(None)`. It is also the first release of two new crates
+that join the lockstep versioning scheme at 0.7.0: `connectrpc-health`
+(the standard gRPC health-checking service) and `connectrpc-reflection`
+(gRPC server reflection).
 
-- **Client streams return terminal server errors from `message()`**.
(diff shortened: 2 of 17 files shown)


iainmcgin:
waiting for the streaming err handling fix and reflection crate before release

rpb-ant (dismissed):

rpb-ant (approved):

PR #162, opened 2026-06-12
router: hide low-level route helpers from rustdoc

### LINKED ISSUE #91: Router::route_* manual routing is stringly typed and surfaces in rustdoc before the codegen path

Yong-yuan-X:
## Summary

- hide all ten stringly-typed `Router::route_*` helpers from rustdoc while keeping them public for generated code
- direct users toward the generated `<Service>Ext::register` API
- replace manual routing examples and references in public documentation
- clarify `Router::with_spec` behavior without hiding the typed API

This focuses on the small scope described in #91: hiding the low-level stringly-typed route helpers from rustdoc and updating the docs to point users toward generated service registration.

Runtime behavior and API compatibility are unchanged.

Fixes #91

## Testing

- `cargo fmt -p connectrpc -- --check`
- `cargo clippy -p connectrpc --all-features --lib -- -D warnings`
- `cargo test -p connectrpc --lib`
- `cargo doc -p connectrpc --all-features --no-deps`
- `cargo check -p multiservice-example --all-features`
- `cargo check -p eliza-example --all-features`

### DIFF
--- a/connectrpc/src/handler.rs
+++ b/connectrpc/src/handler.rs
@@ -344,11 +344,11 @@ where
 /// PreEncoded::<MyResponse>::from_bytes_unchecked(bytes)
 /// ```
 ///
-/// The codegen-emitted route registrations (`route_view_server_stream::<_,
-/// _, Out>`) always pin `Res` because the trait method's stream item is the
-/// *opaque* `impl Encodable<Out>`, which can't be unified against the
-/// `Encodable<Res>` impls. Hand-written `Router` registrations don't hit
-/// this unless they leave the message type open.
+/// Generated server-streaming registrations always pin `Res` because the
+/// trait method's stream item is the *opaque* `impl Encodable<Out>`, which
+/// can't be unified against the `Encodable<Res>` impls. Hand-written
+/// `Router` registrations don't hit this unless they leave the message type
+/// open.
 pub fn streaming_handler_fn<F, Fut, Req, Res, B>(f: F) -> FnStreamingHandler<F>
 where
     F: Fn(RequestContext, Req) -> Fut + Send + Sync + 'static,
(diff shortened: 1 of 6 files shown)


Yong-yuan-X:
Sorry for the close/reopen noise here. While trying to split the generated-code update into a separate commit, I accidentally force-pushed the branch back to `upstream/main`, which closed the PR without merging.

The branch has been restored and the PR is reopened now. I also split the changes into two commits:
- `router: hide low-level route helpers from rustdoc`
- `chore: regenerate generated code`

The second commit only refreshes the checked-in generated outputs needed to address the previous `Check generated code` CI failure.If you’d prefer to keep this PR limited to the docs/API visibility change, I’m happy to adjust the generated-code commit in whatever way you prefer.

iainmcgin (approved):
Thanks for the contribution!

PR #163, opened 2026-06-13
client: require END_STREAM for Connect client-streaming responses

EffortlessSteven:
## What this does

`call_client_stream` accepted a Connect response that held the data message but ended before the END_STREAM envelope, returning it as success with empty trailers. A truncated response was therefore indistinguishable from a complete one. The Connect protocol requires a streaming response to end with an EndStreamResponse envelope.

**Fix:** once the data message is present, `parse_connect_client_stream_envelopes` returns `Err(unavailable)` when END_STREAM was never seen, matching the `ServerStream` Connect EOF path (#146). Complete responses and the no-data and multiple-message rejections are unchanged.

## Verification

Red on `main`, green after: data-then-EOF returns `Ok` unpatched.

| Check | Result |
| --- | --- |
| data + EOF / truncated END_STREAM | `Ok` unpatched, `Err(unavailable)` patched |
| data + END_STREAM `{}` / metadata / error | preserved |
| `fmt`, `clippy`, `cargo test -p connectrpc` | clean, 439 passed |

## Review map

- `connectrpc/src/client/mod.rs`: track whether END_STREAM was seen; a missing terminus after the data message returns `unavailable`. Tests pin missing and truncated END_STREAM, plus the success path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,19 @@ with the [Rust 0.x convention](https://doc.rust-lang.org/cargo/reference/semver.
 breaking changes increment the minor version (0.2 → 0.3), additive changes
 increment the patch version.
 
+## [Unreleased]
+
+### Fixed
+
+- **Connect client-streaming responses require the END_STREAM envelope**
+  ([#163]). A response that ended after its single data message but before
+  the END_STREAM envelope was accepted as a success with empty trailers, so
+  a truncated response was indistinguishable from a complete one. It now
+  returns `Err(unavailable)`, matching the `ServerStream` Connect EOF path
+  ([#140]); complete responses are unchanged.
+
+[#163]: https://github.com/anthropics/connect-rust/pull/163
+
 ## [0.7.0] - 2026-06-10
 
 A breaking release that reworks both ends of the message surface:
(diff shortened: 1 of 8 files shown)


iainmcgin (approved):
[claude code]

Thanks for this — it's a clean, well-scoped fix and the write-up made it easy to verify.

**What we checked**

- Red→green confirmed locally: with the `!saw_end_stream` check neutralized, both new `requires_*` tests fail; with it in place, all 9 `client_stream_response*` tests pass.
- The flag is necessary (the loop has three exits and only the END_STREAM `break` is success), its placement before the `end_stream.error` early-return is harmless, and the error code/message string match the existing `ServerStream` Connect EOF path exactly.
- The doc comment rewrite is accurate; the truncated-envelope test correctly relies on `Envelope::decode_with_limit` returning `None` for a partial body.
- The conservative scope — leaving the empty-body case at `unimplemented` rather than also flipping it to `unavailable` — is the right call for this PR.

**Conformance gap**

You're correct that this isn't covered upstream: `connectrpc/conformance` has no Connect-protocol client test, for any stream type, that sends a data envelope and then closes without the `flags: 2` END_STREAM item. We'll send a PR upstream to add `client-stream/no-end-stream` (and the server-stream/bidi siblings) to `connect_client_unexpected.yaml`.

**Maintainer fixup**

The `Check generated code` failure was pre-existing on `main` (buffa 0.7.1's codegen drifted from the checked-in 0.7.0 output) and unrelated to your change. Pushed a regen commit on top so this merges green; no action needed on your side.

Thanks again for catching this and for the thorough tests.

PR #164, opened 2026-06-14
router: add top-down multi-service registration

### LINKED ISSUE #88: Router: add_service / merge for top-down multi-service registration
Registering multiple services with a `Router` currently reads inside-out:

```rust
let router = Arc::new(svc_b).register(Arc::new(svc_a).register(Router::new()));
```

or, more readably but with intermediate bindings:

```rust
let mut router = Router::new();
router = Arc::new(svc_a).register(router);
router = Arc::new(svc_b).register(router);
```

The `register` extension trait on `Arc<Self>` is a deliberate design — it avoids the visitor-pattern wrapper that tonic generates (`GreeterServer::new(impl)` / `GreeterServer::from_arc(impl)`) — but the call shape inverts the data flow compared to every other framework in the ecosystem:

```rust
// tonic
Server::builder()
    .add_service(GreeterServer::new(greeter))
    .add_service(EchoServer::new(echo))
    .serve(addr).await?;

// connect-go (each service mounts independently on the mux)
mux.Handle(elizav1connect.NewElizaServiceHandler(eliza))
mux.Handle(greetv1connect.NewGreetServiceHandler(greeter))
```

A user arriving from tonic types `Router::new().add_service(...)` first, gets a method-not-found, and has to read the codegen output to discover the `register` extension trait.

## Proposed direction

Add a forwarding builder method on `Router` so registration reads top-down:

```rust
impl Router {
    /// Register a service. Equivalent to `Arc::new(svc).register(self)`.
    pub fn add_service<S, E>(self, svc: Arc<S>) -> Self
    where
        S: ?Sized,
        Arc<S>: ServiceRegister,   // a small trait the codegen `*Ext` traits could implement
    {
        svc.register(self)
    }
}
```

so the call site is:

```rust
let router = Router::new()
    .add_service(Arc::new(eliza))
    .add_service(Arc::new(greeter));
```

This is purely additive — `register(self: Arc<Self>, router)` stays as the underlying mechanism — but it gives the discoverable spelling. While here, also consider promoting the `merge_routers` free function to an inherent `Router::merge(self, other) -> Self` (and an in-place `Router::extend(&mut self, other)` sibling) for the same discoverability reason.

## Scope

Small: one `pub trait` for the bound, one inherent method, doc update, example update.

Yong-yuan-X:
## Summary

- add `Router::add_service` for discoverable, top-down service registration
- generate `ServiceRegister` implementations that forward to the existing service extension traits
- add `Router::merge` and `Router::extend`, with later routes replacing duplicate paths
- update the README, guide, crate documentation, and multi-service example
- regenerate checked-in Connect service code

The existing `FooServiceExt::register` API remains available and continues
to provide the underlying registration behavior, so this change is additive
and backward compatible.

```rust
let router = Router::new()
    .add_service(Arc::new(greeter))
    .add_service(Arc::new(eliza));
```

## Testing

- `cargo +nightly-2026-02-27 fmt --all -- --check`
- `cargo +1.88 check -p connectrpc -p connectrpc-codegen -p multiservice-example --all-features`
- `cargo test -p connectrpc --lib`
- `cargo test -p connectrpc-codegen --lib`
- `cargo test -p connectrpc --no-default-features`
- `cargo check -p multiservice-example --all-features`
- `cargo clippy -p connectrpc -p connectrpc-codegen -p multiservice-example -p connectrpc-health -p connectrpc-reflection -p eliza-example --all-features -- -D warnings`

Closes #88

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,23 @@ increment the patch version.
 
 ## [Unreleased]
 
+### Added
+
+- **Top-down service registration on `Router`** ([#164]). `Router::add_service`
+  registers a generated service from the router outward
+  (`Router::new().add_service(Arc::new(svc))`), the discoverable counterpart to
+  the existing `FooServiceExt::register` extension method, which remains
+  available. New `Router::merge` / `Router::merge_in_place` combine routers, and
+  a new public `ServiceRegister` trait (implemented by codegen) backs
+  `add_service`. Registering or merging a method path that already exists now
+  fails by default so an accidental collision — such as adding the same service
+  twice — surfaces loudly instead of silently shadowing a route: `add_service`,
+  `register`, `merge`, `merge_in_place`, and `merge_routers` panic. Call
+  `Router::allow_overrides` to opt into last-wins replacement across all of
+  them. For assembling routers from dynamic input, `Router::try_merge` /
+  `Router::try_merge_in_place` return a `RouterMergeError` listing the
+  conflicting paths instead of panicking.
+
 ### Fixed
 
 - **Connect client-streaming responses require the END_STREAM envelope**
@@ -20,6 +37,7 @@ increment the patch version.
   ([#140]); complete responses are unchanged.
 
 [#163]: https://github.com/anthropics/connect-rust/pull/163
+[#164]: https://github.com/anthropics/connect-rust/pull/164
 
 ## [0.7.0] - 2026-06-10
 
--- a/README.md
+++ b/README.md
@@ -224,8 +224,7 @@ use axum::{Router, routing::get};
 use connectrpc::Router as ConnectRouter;
 use std::sync::Arc;
 
-let service = Arc::new(MyGreetService);
-let connect = service.register(ConnectRouter::new());
+let connect = ConnectRouter::new().add_service(Arc::new(MyGreetService));
 
 // Plain HTTP liveness probe for `kubectl`'s httpGet style. For the
 // standard gRPC Health protocol (grpc_health_probe, kubelet `grpc:`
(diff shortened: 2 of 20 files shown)


iainmcgin:
[claude code] Thanks for this, @Yong-yuan-X! The `ServiceRegister<Marker>` + `add_service` design is great. I pushed some changes directly to your branch (also rebased onto current `main`, which now includes #162). Here's what changed and why:

### Conflict semantics — fail by default instead of silently overriding

The original `merge` / `extend` silently replaced a route when two routers registered the same path. A silent override is easy to introduce by accident and hard to debug, so I made a path collision fail loudly instead — matching `axum::Router::merge`:

- `add_service`, the generated `register`, the `route_*` helpers, `merge`, `merge_in_place`, and `merge_routers` now reject a path that is already registered.
- `Router::allow_overrides()` opts into last-wins replacement across all of them.
- Distinct services never collide (each path is prefixed with the fully-qualified service name), so this only catches genuine duplicates, such as registering the same service twice.

### Fallible variants

Added `Router::try_merge` / `try_merge_in_place`, which return a new `RouterMergeError` listing the conflicting paths instead of panicking. These are for assembling a router from dynamic input (a plugin list, config-driven service set) where a collision is a condition to handle rather than a programming error.

### Naming

Renamed `extend` to `merge_in_place`. The name `extend` invites the `std::iter::Extend` signature (which takes an iterator), and `merge_in_place` pairs clearly with the owned, chainable `merge`.

### Smaller fixes

- `add_service`'s doc now points the multi-service disambiguation case at the public `FooServiceExt::register(arc, router)` rather than the `#[doc(hidden)]` marker turbofish.
- `RouterMergeError` uses a private field plus a `conflicting_paths()` accessor and `#[non_exhaustive]`, so its shape can evolve without a breaking change.
- Re-exported `merge_routers` at the crate root (every sibling symbol is already there).
- Added a `CHANGELOG` entry, a "combining routers" note in the guide, and tests covering the conflict / override / fallible paths.

Everything is green locally (fmt, `clippy -D warnings`, the full test suite, and docs). Thanks again for the contribution!

iainmcgin (approved):
[claude code] Approving on Iain's behalf (LGTM given in session). The changes made on top of the original contribution are summarized in the comment above; all green locally (fmt, clippy -D warnings, full test suite, docs).

PR #165, opened 2026-06-16
feat(server): add max connection age to BoundServer

### LINKED ISSUE #151: feat(server): add max_connection_age + grace knob to BoundServer
## Summary

The `BoundServer` API has no way to configure a per-connection maximum age. This means servers cannot proactively evict long-lived HTTP/2 connections via GOAWAY, which is recommended for load-balanced deployments (prevents connection pinning, enables zero-downtime rolling restarts).

Tonic provides this via its `MaxConnectionAge` tower layer. We would like parity in connectrpc.rs.

## Proposed API

\`\`\`rust
impl BoundServer {
    /// Send GOAWAY after a connection has been alive for \`duration\`.
    ///
    /// A jitter of up to 10% is applied automatically to prevent thundering-herd
    /// reconnects in large fleets.
    pub fn max_connection_age(mut self, duration: Duration) -> Self { ... }

    /// Grace period after the first GOAWAY before forcibly closing the connection.
    ///
    /// During the grace window the server sends a second GOAWAY, waits for
    /// in-flight streams to complete, then drops the connection. Defaults to
    /// some reasonable value (e.g. 5 s) if not set.
    pub fn max_connection_age_grace(mut self, duration: Duration) -> Self { ... }
}
\`\`\`

## Why this matters

In Kubernetes and other orchestrated environments, services rely on max-connection-age to ensure clients reconnect periodically, distributing load across pod restarts. Without it, the operator must implement this at the service-kit / middleware layer, duplicating logic that properly belongs in the transport.

## Prior art

- tonic: `tonic::transport::Server::max_connection_age` (wraps a tower layer)
- grpc-go: `keepalive.ServerParameters.MaxConnectionAge`
- envoy: `max_connection_duration` in HTTP/2 options

## References

- RFC 9113 §9.3 (HTTP/2 GOAWAY semantics)

Jaksenc:
## Problem

`BoundServer` supports global graceful shutdown, but there is no way to retire long-lived keep-alive connections by age. That makes it hard for servers to periodically drain HTTP/2 connections with GOAWAY while keeping the listener alive for new connections.

## Change

This adds `BoundServer::with_max_connection_age(Duration)` and `BoundServer::with_max_connection_age_grace(Duration)`. Connection age is disabled by default; setting an age uses a five-second grace period unless overridden.

Accepted HTTP/1 and HTTP/2 connections now run through an owned connection lifecycle that composes one shared global shutdown signal with optional per-connection aging. Global shutdown still wins races and drains indefinitely. When max age fires, the connection starts graceful shutdown, waits the configured grace, then drops the connection future to force-close unfinished work.

The configured age gets symmetric +/-10% jitter using the server's `RandomState` plus a per-connection sequence number. The age timer starts when HTTP serving begins, after any TLS accept step has completed.

## Validation

- `cargo test -p connectrpc --features server max_connection_age` passed: 9 focused tests.
- `cargo test -p connectrpc --all-features` passed: 445 unit tests; doctests reported 6 passed and 57 ignored.
- `cargo check -p connectrpc --no-default-features --features server` passed.
- `cargo clippy -p connectrpc --all-features --all-targets -- -D warnings` passed.
- `cargo +nightly-2026-02-27 fmt --all -- --check` passed with the pinned rustfmt toolchain.
- `RUSTDOCFLAGS="-Dwarnings" cargo doc -p connectrpc --all-features --no-deps` passed.
- `git diff --check` passed.

## Risk

The main risk is in the accepted-connection lifecycle, since this changes how global shutdown and connection tasks are coordinated. The tests cover HTTP/2 GOAWAY, listener continuity after one aged connection retires, in-flight completion during age grace, forced close after grace, HTTP/1 keep-alive retirement, global shutdown not being capped by age grace, and the race where global shutdown arrives while an age drain is already in progress.

## Out of scope

This does not add max-age APIs to `Server` or `ServeTls`, does not change generated code, and does not add dependencies.

## Issue linkage

Closes #151

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,6 +138,10 @@ regenerate with this release's toolchain and buffa ≥ 0.7.0.**
 
 ### Added
 
+- **`BoundServer::with_max_connection_age` and
+  `with_max_connection_age_grace`** ([#151]) let the built-in server retire
+  long-lived HTTP/1.1 and HTTP/2 connections gracefully with per-connection
+  jitter while preserving indefinite whole-server graceful shutdown drains.
 - **New `connectrpc-health` crate** ([#128]) — the standard
   `grpc.health.v1.Health` service for connectrpc routers, wire-compatible
   with `grpc_health_probe`, kubelet `grpc:` probes, and gRPC-aware service
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,14 @@ increment the patch version.
   them. For assembling routers from dynamic input, `Router::try_merge` /
   `Router::try_merge_in_place` return a `RouterMergeError` listing the
   conflicting paths instead of panicking.
+- **Maximum connection age for the built-in server** ([#151]).
+  `with_max_connection_age` (on both `Server` and `BoundServer`) retires
+  long-lived HTTP/2 connections by sending a GOAWAY once a connection reaches
+  the configured age, then force-closing after a grace period
+  (`with_max_connection_age_grace`, default 5s); HTTP/1.1 connections have
+  keep-alive disabled instead. A symmetric ±10% jitter is applied per
+  connection to avoid reconnect bursts. Disabled by default; whole-server
+  graceful shutdown still drains in-flight requests indefinitely.
 
 ### Fixed
 
@@ -36,6 +44,7 @@ increment the patch version.
   returns `Err(unavailable)`, matching the `ServerStream` Connect EOF path
   ([#140]); complete responses are unchanged.
 
+[#151]: https://github.com/anthropics/connect-rust/issues/151
 [#163]: https://github.com/anthropics/connect-rust/pull/163
 [#164]: https://github.com/anthropics/connect-rust/pull/164
 
@@ -138,10 +147,6 @@ regenerate with this release's toolchain and buffa ≥ 0.7.0.**
 
 ### Added
 
-- **`BoundServer::with_max_connection_age` and
-  `with_max_connection_age_grace`** ([#151]) let the built-in server retire
-  long-lived HTTP/1.1 and HTTP/2 connections gracefully with per-connection
-  jitter while preserving indefinite whole-server graceful shutdown drains.
 - **New `connectrpc-health` crate** ([#128]) — the standard
   `grpc.health.v1.Health` service for connectrpc routers, wire-compatible
   with `grpc_health_probe`, kubelet `grpc:` probes, and gRPC-aware service
--- a/connectrpc/src/server.rs
+++ b/connectrpc/src/server.rs
@@ -34,9 +34,20 @@
 //! ```
 
 use std::any::Any;
+use std::collections::hash_map::RandomState;
+use std::future::Future;
+use std::hash::BuildHasher;
 use std::net::SocketAddr;
 use std::pin::Pin;
 use std::sync::Arc;
+use std::sync::Mutex;
+use std::sync::Weak;
+use std::sync::atomic::AtomicBool;
+use std::sync::atomic::Ordering;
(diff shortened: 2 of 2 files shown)


Jaksenc:

iainmcgin:
[claude code] Thanks for this, @Jaksenc — really solid work. The per-connection lifecycle state machine and the max-age + grace handling are well thought through, and the test coverage (real h2 GOAWAY, HTTP/1.1 keep-alive retirement, grace-vs-global-shutdown interactions) is excellent. I pushed some changes directly to your branch (also rebased onto current `main`). Here's what changed and why:

### Replaced the hand-rolled shutdown broadcast with `tokio::sync::watch`

The custom `GlobalShutdown` / `GlobalShutdownWait` primitive (an `AtomicBool` plus a `Mutex<Vec<Weak<Mutex<Option<Waker>>>>>`) was traced and found correct — no lost-wakeup or missed-shutdown race. But it reimplements an audited primitive, and for a published crate the bar for keeping ~90 lines of hand-written `Waker` management is high. Swapping it for `watch::channel(false)` (already in our dependency tree, zero new deps) removes that maintenance/audit burden, turns the O(n²) waiter-drain into O(1) register/deregister, and eliminates a latent `wake()`-while-holding-the-lock footgun. Each connection holds a receiver and awaits `wait_for(|fired| *fired)`; the accept loop flips the value on shutdown. Behaviour is unchanged and covered by the existing tests.

One nice side effect: on a fatal accept error the loop drops the sender, which `wait_for` surfaces as a closed channel, so detached connections now drain gracefully (GOAWAY) instead of being abandoned mid-flight.

### Added `Server` parity for the new setters

`with_max_connection_age` / `with_max_connection_age_grace` were only on `BoundServer`, so the one-step `Server::new(router).serve(addr)` path couldn't enable them. Every other transport knob (`with_tls`, `with_http1_keep_alive`) is mirrored on both types, so I added them to `Server` and threaded them through `serve`, matching that convention.

### Smaller fixes

- **Zero duration**: `with_max_connection_age(Duration::ZERO)` retires every connection on arrival. Added a `debug_assert!` and a doc caveat so the misconfiguration surfaces in development.
- **Grace without max-age**: setting only the grace is a no-op; added a `tracing::debug!` and a doc note so it isn't a silent surprise.
- **Docs**: noted that the age path sends a GOAWAY (matching the global-shutdown docs), added a module-level "Maximum Connection Age" overview, and moved the CHANGELOG entry from the already-released `0.7.0` section to `[Unreleased]` (with the issue link defined).
- **Tests**: added a `Server`-parity test and two tests pinning the shutdown future's resolution on both the signal and the sender-drop (fatal-accept) paths.
- Removed a now-redundant manual `Unpin` impl (the struct is auto-`Unpin`).

Everything is green locally (fmt, `clippy --all-targets -D warnings`, the full test suite, and docs). Thanks again for the contribution!

iainmcgin:
[claude code] Follow-up: changed the zero-`max_age` guard from `debug_assert!` to a real `assert!`, so an illegal zero duration is rejected in release builds too (a `debug_assert` would let a production build silently turn into a connection shredder). `Duration` is unsigned so zero is the only illegal value; `with_max_connection_age_grace(0)` stays legal (GOAWAY-then-close, no drain). Opened #173 to bring the other config setters (TLS handshake timeout, deadline timeouts) into line with the same reject-illegal-values convention.

iainmcgin (approved):
[claude code] Approving on Iain's behalf (go-ahead given in session). Replaced the hand-rolled shutdown primitive with tokio::sync::watch, added Server parity, reject-zero validation, and docs/tests; both review agents + a dedicated concurrency re-review are clean, full suite green locally. Follow-up consistency work tracked in #173.

PR #167, opened 2026-06-18
client: emit Connect Unary-Get query parameters in spec-recommended order

iainmcgin:
## What this does

The Connect protocol's [Query-Get ABNF rule](https://connectrpc.com/docs/protocol#unary-get-request) fixes the Unary-Get query-parameter order as `connect`, `base64`, `compression`, `encoding`, `message` and recommends clients follow it ("Clients should order parameters as shown in the Query-Get rule above to maximize hit rates on shared caches"): the variable-length `message` comes last, so the URL prefix is stable. `connect-go` emits this order, and the upstream conformance suite recently added a reference-server check for it.

Our previous emission was `connect, encoding, message, base64, compression`, which fails that check (4 permutations under `Connect with GET/**` against conformance HEAD).

Query assembly is factored into a private `build_connect_get_query` helper so the order can be unit-tested directly, with capacity pre-sized to the exact final length. Servers must accept any order, so this is not a wire-compatibility change.

## Verification

- Unit: 4 new tests pin the exact strings for the reachable shapes (proto, uncompressed JSON, compressed proto, compressed JSON), plus a rank-order helper that re-derives the order from the same `[connect, base64, compression, encoding, message]` table the conformance check uses.
- Conformance HEAD (`connectrpc/conformance@148f555` + the `client-stream/no-end-stream` addition from connectrpc/conformance#1104), client-connect-only config: `2584 passed, 0 failed` (was `2580 passed, 4 failed` before this change).
- `task lint`, `task test` clean.

## Review map

- `connectrpc/src/client/mod.rs`: `call_unary_get` doc comment updated; query construction moved into `build_connect_get_query`; tests under the `call_unary_get query encoding` section.
- `CHANGELOG.md`: `### Changed` entry under Unreleased.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,17 @@ increment the patch version.
 
 ## [Unreleased]
 
+### Changed
+
+- **Connect Unary-Get query parameters are emitted in the spec-recommended
+  order** ([#167]): `connect`, `base64`, `compression`, `encoding`, `message`.
+  Servers must accept any order, so this is not a wire-compatibility change;
+  the recommended order keeps the variable-length `message` last so the URL
+  prefix is stable for shared caches. Aligns with the order check added to
+  the upstream conformance suite.
+
+[#167]: https://github.com/anthropics/connect-rust/pull/167
+
 ### Fixed
 
 - **Connect client-streaming responses require the END_STREAM envelope**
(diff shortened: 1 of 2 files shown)


aaomidi (dismissed):
:shipit:

aaomidi (approved):
:shipit:

PR #169, opened 2026-06-18
client: cap server-supplied trailer metadata to avoid HeaderMap panic

iainmcgin:
## What

The client builds an `http::HeaderMap` from server-supplied trailer
metadata in three places, each appending one entry per field with the
panicking `HeaderMap::append`:

- the gRPC-Web trailer frame parser (`parse_grpc_web_trailer_frame_with_compression`), and
- both Connect end-stream paths, which deserialize a JSON `metadata` map (`HashMap<String, Vec<String>>`) and copy it in.

`HeaderMap` has a hard ceiling on the number of stored entries
(`MAX_SIZE = 1 << 15 = 32_768`), and `append` panics with
`"size overflows MAX_SIZE"` once that ceiling is crossed.

## Why it matters

All three sources are attacker-controlled and only bounded by *bytes*,
not by entry count:

- The gRPC-Web parser caps the trailer frame at 1 MiB, but more than
  32,768 short, distinct trailer lines (`h0:`, `h1:`, … `h39999:`) fit
  in roughly 300 KiB.
- The Connect `metadata` map comes straight from the JSON end-stream
  frame; tens of thousands of distinct keys are a few hundred KB of JSON.

In each case parsing the response panics the RPC task. Because the
metadata rides in the body rather than in real HTTP headers, the
transport's own header-count limits never see it. The Connect path is
the default codec, so a typical client is exposed; the gRPC-Web path
covers the streaming and wasm clients.

## Fix

Switch all three sites from `append` to the fallible `try_append`,
stopping once the map is full. The two Connect paths now share a single
`append_metadata_capped` helper so they cannot drift:

```rust
if trailers.try_append(name, value).is_err() {
    break;
}
```

This rejects nothing legitimate — real trailers are a handful of names,
far below the ceiling — and turns a panic on oversized input into a
graceful stop that preserves whatever fit (an early `grpc-status`
survives).

## Testing

- `test_parse_grpc_web_trailer_floods_do_not_panic`: a 40,000-name
  gRPC-Web frame (~300 KiB, under the byte cap) that panicked before the
  change and now parses with the map capped at `MAX_SIZE`.
- `test_append_metadata_capped_does_not_panic_on_flood`: the Connect
  equivalent over a 40,000-key `metadata` map.
- `test_append_metadata_capped_copies_entries`: confirms normal metadata
  (including multi-valued names) is copied through unchanged.

Existing trailer-parser tests continue to pass.

### DIFF
--- a/.github/workflows/cla.yaml
+++ b/.github/workflows/cla.yaml
@@ -19,14 +19,17 @@ jobs:
     timeout-minutes: 5
     steps:
       - name: "CLA Assistant"
-        if: (github.event.comment.body == 'recheck' || github.event.comment.body == 'I have read the CLA Document and I hereby sign the CLA') || github.event_name == 'pull_request_target'
+        # contains() rather than == so the action also fires when the sign
+        # phrase or `recheck` is accompanied by a small amount of extra text;
+        # the action itself enforces the precise matching policy.
+        if: github.event_name == 'pull_request_target' || contains(github.event.comment.body, 'recheck') || contains(github.event.comment.body, 'I have read the CLA Document and I hereby sign the CLA')
         # Upstream contributor-assistant/github-action was archived 2026-03-23
         # still on Node 20 (deprecated 2026-06-02). This fork bumps to Node 24
         # and adds: an impersonation guard (PR opener must be an author or
         # co-author of at least one commit), Co-authored-by trailer support,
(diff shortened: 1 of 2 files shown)


aaomidi (approved):
:shipit:

PR #172, opened 2026-06-19
connectrpc: add optional `json` feature for proto-only builds

macalinao:
## Motivation

Adding `serde` to **every** protobuf-generated type is a real compile-time tax. The Connect JSON codec requires `serde::Serialize`/`Deserialize` on every message type, so the code generator derives them by default — and for a workspace with many `.proto` files, those derives (and the `serde_derive` proc-macro expansion + monomorphized `serde_json` paths) add up across hundreds of generated structs. Crates that only ever speak **binary proto** pay that cost for nothing.

This adds a way to opt out: generate without serde derives, and compile against a `connectrpc` that doesn't demand them.

## What this does

A default-on **`json` cargo feature** on `connectrpc`. With it disabled (`default-features = false`):

- Message-type bounds relax from `Message + Serialize`/`DeserializeOwned` to just `Message`, via two new public marker traits **`JsonSerialize`/`JsonDeserialize`** (serde supertraits when `json` is on, empty blanket bounds when off). So types generated with the existing codegen **`no_json`** option (which drops the serde derives) compile against the runtime.
- JSON message encode/decode is gated at a **single point** — `codec::encode_json`/`decode_json` — which returns `Unimplemented` when the feature is off. A JSON request to a proto-only server gets a clean Connect error; content-type negotiation stays codec-agnostic.
- `connectrpc`'s `buffa` dependency is pinned directly (not `workspace = true`) so the feature truly gates `buffa/json` — a proto-only build resolves `buffa` with **no features**, shedding its JSON runtime too.

The Connect **error** and **end-of-stream** wire frames are always JSON per the protocol spec (they use connectrpc's own structs), so `serde`/`serde_json` stay required dependencies — only *message-type* JSON is gated.

## Proto-only recipe

```toml
# generate without serde derives: codegen `no_json`, or connectrpc-build's
# .generate_json(false)
connectrpc = { version = "0.7", default-features = false, features = ["server"] }
```

Verified `cargo tree -p connectrpc --no-default-features` resolves `buffa` with no features; with `json` on it shows `buffa json`.

## Implementation notes

- ~50 trait-bound sites swapped to the marker traits; all `CodecFormat::Json` message arms route through `encode_json`/`decode_json` (no per-site `#[cfg]`). One deliberate exception: the client's `decode_response_view` keeps an explicit gate to preserve its `internal` error code (response-decode semantics ≠ `decode_json`'s `invalid_argument`).
- Docs: proto-only section in the guide + README; `connectrpc-build::generate_json` and the codegen `no_json` option docs updated.

## Testing

- `cargo test -p connectrpc` (json on): **403 passed**.
- `cargo test -p connectrpc --no-default-features` (proto-only): **349 passed**, including new `#[cfg(not(feature = "json"))]` assertions that every fallback (response / payload / handler / stream_message / client) returns `Unimplemented`, plus a marker-empty-bound compile assertion.
- Clippy, nightly `fmt`, and `cargo doc --all-features` clean in both configs.

> Reviewed by the repo's `rust-code-reviewer` and `rust-api-ergonomics-reviewer` (no Critical/High/Medium outstanding).

## Follow-ups (out of scope)

- Workspace-wide `buffa`/`buffa-types` de-unification so `connectrpc-health` / `connectrpc-reflection` and other members can also shed `buffa/json` when proto-only.
- Matching `json` features on `connectrpc-health` / `connectrpc-reflection`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,9 +20,15 @@ increment the patch version.
   message-type bounds to just `buffa::Message` via the new
   `JsonSerialize`/`JsonDeserialize` marker traits, so message types
   generated with the codegen `no_json` option (no serde derives) compile
-  against the runtime. A JSON request to a proto-only server returns
-  `Unimplemented`. The Connect error/end-stream wire format is always JSON per
-  spec, so `serde`/`serde_json` remain required dependencies. See the
+  against the runtime. A proto-only server declines JSON at content
+  negotiation — `application/json` / `application/connect+json` (and the
+  Connect GET `encoding=json` parameter) return HTTP 415, and
+  `application/grpc+json` / `application/grpc-web+json` return a gRPC error
+  status — with message-level encode/decode returning `Unimplemented` as a
+  backstop; the client's `ClientConfig::json` selector is removed from the API
+  in that build. The Connect error/end-stream wire format
+  is always JSON per spec, so `serde`/`serde_json` remain required
+  dependencies. See the
   [proto-only build guide](docs/guide.md#proto-only-no-json-builds).
 
 [#172]: https://github.com/anthropics/connect-rust/pull/172
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,23 @@ increment the patch version.
 
 ## [Unreleased]
 
+### Added
+
+- **Optional `json` cargo feature for proto-only builds** ([#172]). The
+  Connect JSON codec requires `serde::Serialize`/`Deserialize` on every
+  message type, so the code generator derives them by default — pure cost for
+  crates that only speak binary proto. The new default-on `json` feature, when
+  disabled (`connectrpc = { default-features = false }`), relaxes the runtime's
+  message-type bounds to just `buffa::Message` via the new
+  `JsonSerialize`/`JsonDeserialize` marker traits, so message types
+  generated with the codegen `no_json` option (no serde derives) compile
+  against the runtime. A JSON request to a proto-only server returns
+  `Unimplemented`. The Connect error/end-stream wire format is always JSON per
+  spec, so `serde`/`serde_json` remain required dependencies. See the
+  [proto-only build guide](docs/guide.md#proto-only-no-json-builds).
+
+[#172]: https://github.com/anthropics/connect-rust/pull/172
+
 ### Fixed
 
 - **Connect client-streaming responses require the END_STREAM envelope**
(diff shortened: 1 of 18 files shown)


macalinao:

iainmcgin:
[claude code] Thanks for this — the core design (default-on `json` feature, the `JsonSerialize`/`JsonDeserialize` marker traits, single-point codec gating, and the direct `buffa` pin) is clean and well-reasoned. I reviewed it with the repo's two Rust review agents and verified it end to end (both feature configs build, clippy `-D warnings`, `cargo doc`, and the full Connect/gRPC/gRPC-Web conformance suites all pass). I pushed one follow-up commit (`56a860c`) with maintainer changes rather than asking you to round-trip them; happy to walk through anything.

What the follow-up commit adds, and why:

**Reject JSON at content negotiation (not just at decode).** The original PR returned `Unimplemented` from `encode_json`/`decode_json`, which made a JSON request to a proto-only server fail *after* the request was accepted and decompressed. The Connect protocol requires a server that doesn't support a message codec to respond with **HTTP 415 Unsupported Media Type**, up front. So in a proto-only build the JSON arms of `Protocol::detect_from_content_type`, `grpc_subtype_to_codec`, `CodecFormat::from_content_type`/`from_codec`/`is_streaming_content_type` are now `#[cfg(feature = "json")]`-gated: `application/json` / `application/connect+json` (and the Connect GET `encoding=json` parameter) return 415, and `application/grpc+json` / `application/grpc-web+json` return a gRPC error status. The message-level `Unimplemented` stays as a defense-in-depth backstop. This deliberately supersedes the original "keep negotiation codec-agnostic for a precise Unimplemented" note — the spec wants 415 at negotiation, and this matches connect-go (which derives its accepted content types from the registered codecs).

**`Accept-Post` on 415.** 415 responses now advertise the content types the server accepts (proto-only builds omit the JSON media types), matching connect-go.

**Unsupported-verb handling.** Any non-GET/POST verb now returns **405 with an `Allow` header** (POST, plus GET for idempotent unary methods), with unknown paths returning 404 — fixing a bodyless OPTIONS/HEAD previously being misreported as 415.

**Client selector redaction.** `ClientConfig::json()` is now `#[cfg(feature = "json")]`, so JSON can't be selected in a proto-only build.

**Smaller items.** The `buffa` pin keeps `std` explicitly (the feature should gate only `buffa/json`); the `json` feature-table row, the cargo feature-unification caveat, and the "disabling defaults also drops compression" note were added to the README/guide; and a few doc comments were corrected.

Two notes:
- Disabling default features to drop `json` also drops `gzip`/`zstd`/`streaming`, so the proto-only recipes re-list them. And proto-only mode only holds if every `connectrpc` consumer in the dependency graph disables `json` (additive features) — both are documented now.
- The gRPC unsupported-codec status is currently `internal`; the conformance suite explicitly allows `internal`/`unimplemented`/`unknown` here, so it isn't a bug, but for consistency with the unsupported-compression path I filed #180 to move it to `unimplemented` separately.

Behavior change worth flagging for downstreams: OPTIONS/HEAD requests now return 405 (previously 415) — more correct, and CORS preflight is expected to be handled by middleware.

iainmcgin (approved):

PR #183, opened 2026-06-20
grpc: return unimplemented for unsupported message codec

### LINKED ISSUE #180: gRPC: use `unimplemented` (not `internal`) for an unsupported message codec
## Summary

When a gRPC or gRPC-Web request arrives with a valid `application/grpc` (or `application/grpc-web`) prefix but an **unsupported message codec** — for example `application/grpc+thrift`, or `application/grpc+json` on a build without the `json` feature — the server rejects it with a `grpc-status` error whose code is `internal`:

```rust
// connectrpc/src/service.rs
let err = ConnectError::internal("unsupported content type");
```

For the analogous failure on the **compression** axis (the client asks for an encoding the server does not offer, e.g. `grpc-encoding: foo`), the server already returns `unimplemented` (see `connectrpc/src/compression.rs`). Using `internal` for an unsupported codec but `unimplemented` for an unsupported compression is inconsistent, and `unimplemented` ("the service does not support this codec") is the more accurate code than `internal` ("a server invariant was broken").

## Proposed change

Switch the unsupported-codec gRPC / gRPC-Web rejection from `internal` to `unimplemented`, matching the compression sibling.

## Why this is low priority (not a bug)

- The gRPC HTTP/2 spec is silent on which status to use for a valid `application/grpc` prefix with an unsupported codec. It only mandates HTTP 415 when the content type does **not** begin with `application/grpc`.
- The Connect conformance suite explicitly treats this code as unspecified. In `grpc_server_unexpected.yaml` the `unexpected-codec` case lists `otherAllowedErrorCodes: [CODE_INTERNAL, CODE_UNIMPLEMENTED]` with the comment *"Not actually specified what error code to use ... There may be an argument for unimplemented, too."*

So the current `internal` behavior is conformance-valid; this is a consistency and semantic-accuracy improvement, not a correctness fix.

## Scope

- `connectrpc/src/service.rs`: the two `ConnectError::internal("unsupported content type")` sites on the gRPC and gRPC-Web branches of the unknown-protocol path.
- Affects `application/grpc+<unsupported>` and `application/grpc-web+<unsupported>`.

## Context

Surfaced while reviewing the optional-`json`-feature work: in a proto-only build, `application/grpc+json` routes through this same path, which made the existing `internal` choice visible. The codec-rejection code is otherwise unchanged by that work.

iainmcgin:
## What

When a gRPC or gRPC-Web request arrives with a valid `application/grpc` (or `application/grpc-web`) prefix but an **unsupported message codec** — for example `application/grpc+thrift`, or `application/grpc+json` on a build without the `json` feature — the server now rejects it with grpc-status `unimplemented` (12) instead of `internal` (13).

## Why

`unimplemented` ("the service does not support this codec") is the more accurate code than `internal` ("a server invariant was broken"). It also brings the codec axis into parity with the compression axis, which already returns `unimplemented` for an unsupported `grpc-encoding`.

The gRPC HTTP/2 spec is silent on which status to use for a valid `application/grpc` prefix with an unsupported codec, and the Connect conformance suite explicitly accepts either `internal` or `unimplemented` for this case (`grpc_server_unexpected.yaml`, `unexpected-codec`). So this is a consistency and semantic-accuracy improvement, not a correctness fix.

## Changes

- `connectrpc/src/service.rs`: the two `ConnectError::internal("unsupported content type")` sites on the gRPC and gRPC-Web branches of the unknown-protocol/codec path switch to `ConnectError::unimplemented(...)`. Updated the surrounding comment to name the code.
- New colocated unit test asserting grpc-status 12 for an unsupported codec on both the gRPC branch (status echoed in response headers) and the gRPC-Web branch (status carried in the trailer body frame).
- `CHANGELOG.md`: `### Changed` entry noting the observable 13 → 12 status-code change for clients that branch on grpc-status.

## Validation

`task lint` (clippy `-D warnings` + fmt check) and `task test` both pass; the new test passes in the default and `--no-default-features` (proto-only) builds.

Closes #180

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,10 +62,22 @@ increment the patch version.
   returns `Err(unavailable)`, matching the `ServerStream` Connect EOF path
   ([#140]); complete responses are unchanged.
 
+### Changed
+
+- **Unsupported gRPC/gRPC-Web message codecs return `unimplemented`**
+  ([#180]). A request with a valid `application/grpc` / `application/grpc-web`
+  prefix but a codec the server does not speak (for example
+  `application/grpc+thrift`, or `application/grpc+json` in a proto-only build)
+  now returns grpc-status `unimplemented` (12) instead of `internal` (13).
+  This matches the compression axis, which already returns `unimplemented` for
+  an unsupported `grpc-encoding`. Clients that branch on grpc-status will
+  observe 13 → 12 for this case on upgrade.
+
 [#151]: https://github.com/anthropics/connect-rust/issues/151
 [#163]: https://github.com/anthropics/connect-rust/pull/163
 [#164]: https://github.com/anthropics/connect-rust/pull/164
 [#172]: https://github.com/anthropics/connect-rust/pull/172
+[#180]: https://github.com/anthropics/connect-rust/issues/180
 
 ## [0.7.0] - 2026-06-10
 
--- a/connectrpc/src/service.rs
+++ b/connectrpc/src/service.rs
@@ -1473,8 +1473,10 @@ where
 
     // Content type didn't resolve to a known protocol+codec. A gRPC/gRPC-Web
     // prefix with an unsupported codec (e.g. application/grpc+thrift, or
-    // application/grpc+json in a proto-only build) returns a gRPC-style error so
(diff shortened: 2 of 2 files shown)


aaomidi (approved):
:shipit:

PR #187, opened 2026-06-20
client: bound connection establishment (TCP, TLS, HTTP/2 handshake)

### LINKED ISSUE #137: client: bound connection establishment (TCP, TLS, HTTP/2 handshake)
`HttpClientBuilder::connect_timeout` covers only the TCP connect; DNS resolution and the TLS handshake are unbounded unless the caller sets a per-call timeout. The shared `Http2Connection` transport has no connect-timeout knob at all: the TCP connect, TLS handshake, and HTTP/2 preface (`connectrpc/src/client/http2.rs`) all run without a time bound inside the reconnect state machine, so a server that accepts the connection but stalls during the handshake also stalls `poll_ready` for every other caller sharing that transport.

Proposed work:

- Add a handshake timeout covering TLS and the HTTP/2 preface on both client transports (mirroring the server-side TLS handshake timeout).
- Give `Http2Connection` a connect-timeout option.
- Document that `CallOptions::with_timeout` remains the end-to-end bound.

iainmcgin:
## What

Bounds client connection establishment so a hung or slow server can't make a client wait indefinitely while a connection comes up.

- **`Http2ConnectionBuilder`** (new, via `Http2Connection::builder()`):
  - `connect_timeout(Duration)` — bounds the TCP connect (applied to the built-in connector via hyper's `set_connect_timeout`).
  - `handshake_timeout(Duration)` — bounds the TLS handshake plus the HTTP/2 preface, via `tokio::time::timeout` around the post-connect work in the reconnect path.
  - Terminals `lazy_plaintext` / `connect_plaintext` / `lazy_tls` / `connect_tls`. The existing `Http2Connection` constructors now delegate to the builder with no bounds set.
- **`HttpClientBuilder::handshake_timeout(Duration)`** (new) — bounds the whole connector establishment (TCP + TLS) via a small `TimeoutConnector` wrapper around the connector.

## Why

`HttpClientBuilder::connect_timeout` (PR #117) bounded only the TCP connect; the TLS handshake was unbounded. `Http2Connection` had no connect bound at all, so a server that accepted the TCP connection but stalled the handshake stalled `poll_ready` for every caller sharing the transport.

## Behaviour and scope

- Both bounds default to unset, so existing callers are unaffected.
- `Http2Connection`'s `handshake_timeout` covers the TLS handshake + h2 preface (TCP is covered separately by `connect_timeout`). On plaintext h2c, hyper's handshake resolves locally without waiting for the server, so the bound there only catches slow local setup — a stalled cleartext server is caught by a per-request timeout. This is documented on the setter.
- `HttpClient`'s `handshake_timeout` bounds TCP + TLS as one budget; the h2 preface runs inside hyper's connection pool and is not separately observable, so it is not covered there. Documented on the setter.
- `CallOptions::with_timeout` remains the end-to-end per-request bound.

## Testing

New colocated unit tests: builder defaults/setters; `connect_timeout` fast-fails against an unroutable TEST-NET-1 address; `handshake_timeout` fires against a listener that accepts TCP but stalls the TLS handshake (both `Http2Connection` and `HttpClient`); and a happy-path test that a generous bound doesn't interfere with a real h2c server. `task lint`, `task test` (workspace, all features), clippy `-D warnings`, and the broken-intra-doc-links doc build all pass.

Closes #137

### DIFF
--- a/connectrpc/src/client/http2.rs
+++ b/connectrpc/src/client/http2.rs
@@ -34,6 +34,7 @@ use std::future::Future;
 use std::pin::Pin;
 use std::task::Context;
 use std::task::Poll;
+use std::time::Duration;
 
 use http::Request;
 use http::Response;
@@ -242,6 +243,20 @@ fn require_http_scheme(uri: &Uri) -> Result<(), ConnectError> {
 }
 
 impl Http2Connection {
+    /// Returns a builder for bounding connection establishment (TCP connect,
+    /// TLS handshake, HTTP/2 preface) before choosing a transport flavour.
+    ///
+    /// The constructors on `Http2Connection` are shortcuts for the builder
+    /// with no bounds set, so `Http2Connection::lazy_plaintext(uri)` is exactly
+    /// `Http2Connection::builder().lazy_plaintext(uri)`. Reach for the builder
+    /// when you want to bound a stalled connect or handshake — see
+    /// [`Http2ConnectionBuilder::connect_timeout`] and
+    /// [`Http2ConnectionBuilder::handshake_timeout`].
+    #[must_use]
+    pub fn builder() -> Http2ConnectionBuilder {
+        Http2ConnectionBuilder::default()
+    }
+
     /// Create a **plaintext** h2c connection that establishes lazily on
     /// first `poll_ready`. Only for `http://` URIs.
     ///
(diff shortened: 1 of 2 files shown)


iainmcgin:
closing in favor of #197

PR #192, opened 2026-06-22
Reject malformed Connect END_STREAM JSON

fallintoplace:
## Summary

- Reject malformed Connect END_STREAM JSON as an internal protocol error instead of treating it as an empty successful terminator.
- Share END_STREAM parsing, trailer conversion, and error conversion between the server-stream and client-stream response paths.
- Add regression coverage for malformed END_STREAM JSON in both paths.

## Why

The client used `serde_json::from_slice(...).unwrap_or_default()` when parsing Connect END_STREAM payloads. Invalid JSON therefore became the default successful end-stream shape, so a malformed protocol terminator could be reported as a clean RPC completion.

## Testing

- `cargo test -p connectrpc malformed_end_stream_json`
- `cargo test -p connectrpc`
- `cargo test --workspace`
- `cargo clippy -p connectrpc --all-targets --all-features -- -D warnings`
- `git diff --check`

### DIFF
--- a/connectrpc/src/client/mod.rs
+++ b/connectrpc/src/client/mod.rs
@@ -2328,36 +2328,15 @@ where
             Err(e) => return e.into(),
         };
 
-        let end_stream: ClientEndStreamResponse =
-            serde_json::from_slice(&end_stream_data).unwrap_or_default();
+        let end_stream = match parse_connect_end_stream(&end_stream_data) {
+            Ok(end_stream) => end_stream,
+            Err(e) => return e.into(),
+        };
 
-        let trailers = end_stream.metadata.map(|metadata| {
-            let mut trailers = http::HeaderMap::new();
-            for (name, values) in metadata {
-                for value in values {
-                    if let (Ok(name), Ok(value)) = (
-                        http::header::HeaderName::from_bytes(name.as_bytes()),
-                        http::header::HeaderValue::from_str(&value),
-                    ) {
-                        trailers.append(name, value);
-                    }
-                }
-            }
-            trailers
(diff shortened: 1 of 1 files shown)


fallintoplace:

iainmcgin (approved):
[claude code] Thanks — the fix is correct and the regression tests are well-targeted. I pushed one maintainer commit (d55d2da) that merges `main` and resolves the conflict in `client/mod.rs`:

- **Dropped `metadata_to_trailers`** and kept `main`'s `append_metadata_capped` at both call sites. The extracted helper carried the pre-#169 uncapped `HeaderMap::append` loop, which panics at the map's ~32k-entry ceiling — a server can hit that with a few hundred KB of END_STREAM JSON. #169 replaced both sites with `try_append`-and-stop; the resolution preserves that guard while keeping your `parse_connect_end_stream` / `end_stream_error_to_connect_error` extractions.
- **Removed the `Default` derive on `ClientEndStreamResponse`** — it existed only for the `unwrap_or_default()` this PR removes, and leaving it would let a future `.unwrap_or_default()` compile and silently reintroduce the bug.
- **Added a `CHANGELOG.md` entry** under `[Unreleased] / Fixed` alongside its #163/#168 siblings, plus doc comments on the two new helpers.

One pre-existing asymmetry surfaced during review (not a regression here, so not addressed in this PR): the server-stream parse-error path at `process_end_stream` doesn't attach `self.headers` to the returned `ConnectError`, while the client-stream path does. Worth a follow-up to make them uniform.

LGTM with the above applied.

PR #193, opened 2026-06-22
Clamp client timeouts before deadlines

fallintoplace:
## Summary

- Clamp client request timeouts to the protocol timeout header budget before creating local deadlines.
- Replace direct `Instant::now() + timeout` calls with checked deadline construction.
- Keep timeout header formatting and local deadline enforcement tied to the same encoded timeout value.

## Why

`CallOptions::with_timeout` and `ClientConfig::with_default_timeout` accept arbitrary `Duration`s. Very large durations could overflow `std::time::Instant` when the client computed `Instant::now() + timeout`, panicking the caller before the request was sent.

## Testing

- `cargo test -p connectrpc timeout`
- `cargo test -p connectrpc`
- `cargo clippy -p connectrpc --all-targets --all-features -- -D warnings`
- `cargo test --workspace`
- `git diff --check`

### DIFF
--- a/connectrpc/src/client/mod.rs
+++ b/connectrpc/src/client/mod.rs
@@ -1029,6 +1029,89 @@ fn merge_headers(config_defaults: &http::HeaderMap, options: http::HeaderMap) ->
     merged
 }
 
+const CONNECT_TIMEOUT_MAX_MILLIS: u64 = 9_999_999_999;
+const GRPC_TIMEOUT_MAX_SECONDS: u64 = 99_999_999;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum EncodedTimeout {
+    Connect {
+        millis: u64,
+    },
+    Grpc {
+        value: u64,
+        unit: char,
+        duration: Duration,
+    },
+}
+
+impl EncodedTimeout {
+    fn duration(self) -> Duration {
+        match self {
+            Self::Connect { millis } => Duration::from_millis(millis),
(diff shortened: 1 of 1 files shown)


iainmcgin (approved):
[claude code] LGTM — the panic fix is correct and complete: every client RPC entry point now derives both the timeout header and the local deadline from a single clamped `EncodedTimeout`, `Instant::now().checked_add()` guards the addition, and the regression tests pin the exact `Duration::MAX` panic path. Thanks for the thorough test coverage.

Two non-blocking notes (happy to take these as a follow-up or fold them in if you'd like):

- **Sub-millisecond Connect timeouts now floor to an immediate local deadline.** `Duration::from_micros(500)` → `as_millis() == 0` → `client_deadline` returns `Some(now)` → `DeadlineExceeded` before any I/O. Pre-PR the header was already `"0"` (so a compliant server rejected it anyway), so the outcome against compliant servers is unchanged — but worth either `millis = as_millis().max(1)` when `timeout > Duration::ZERO`, or a doc note on `with_timeout` that Connect's floor is 1 ms.
- **`with_timeout` / `with_default_timeout` rustdoc** could mention that the value is clamped to the protocol wire-format ceiling (so `Duration::MAX` is now a valid "effectively unbounded" sentinel).

Minor cleanup opportunity (also non-blocking): the carried-over `#[allow(clippy::manual_is_multiple_of)]` can be dropped now that MSRV ≥ 1.87.

PR #194, opened 2026-06-23
Make generated client feature gate configurable

### LINKED ISSUE #181: connectrpc-codegen: make the `client` feature-gate name configurable
## Summary

`connectrpc-codegen` can gate the generated client-side items behind a Cargo feature (the opt-in `gate_client_feature` option), but the feature **name** it emits is hardcoded to `"client"`:

```rust
// connectrpc-codegen/src/codegen.rs
quote! { #[cfg(feature = "client")] }
```

An integrator whose own crate gates its client surface behind a differently-named feature (e.g. `grpc-client`, `net`, `transport`) cannot make the generated `#[cfg(feature = "...")]` match. Their options are to adopt the literal `client` feature name or to post-process the generated output.

## Proposal

Let the client gate name be configured, rather than fixed. Concretely, either:

- accept a name on the existing opt-in — e.g. `gate_client_feature(Some("grpc-client"))` / a companion `client_feature_name(&str)` — defaulting to `"client"`, or
- thread it through a small "feature gate names" config, mirroring buffa.

The default stays `"client"`, so this is purely additive.

## Why (consistency with buffa)

buffa already supports customizable feature-gate names for the impls it generates, via `gate_impls_on_crate_features` + `FeatureGateNames` (see `buffa-codegen`). This issue brings the one feature gate `connectrpc-codegen` emits in line with that philosophy, so generated code can match the feature names integrators actually use.

## Notes / priority

Low priority. connectrpc service stubs usually live in the top layer of a service build rather than in a published library, so feature-name matching matters less here than it does for buffa's message types. This is the only feature gate `connectrpc-codegen` emits — it does **not** emit a `json` gate (the `JsonSerialize`/`JsonDeserialize` marker traits keep generated stubs codec-agnostic), so there is nothing else to make configurable on this axis.

A separate issue covers surfacing buffa's message-type gating options through `connectrpc-build`.

Jaksenc:
## Problem

`protoc-gen-connect-rust` can already gate generated client items behind a Cargo feature, but the feature name is fixed to `client`. Crates that expose their generated client surface under a different feature name cannot use the existing gate directly.

## Change

- Add `Options::client_feature_name`, defaulting to `client`, and thread it into generated client cfg attributes.
- Accept `gate_client_feature=<name>` in the protoc plugin while preserving the existing bare `gate_client_feature` behavior.
- Add `connectrpc_build::Config::client_feature_name(...)` as the build wrapper companion API.
- Reject empty custom feature names and keep the generated cfg as a string literal so names like `grpc-client` work.
- Document the additive option in the changelog.

This keeps the existing opt-in shape intact: `gate_client_feature` still enables gating, and the default gate remains `client`.

## Validation

Using `protoc` v33.5:

- `cargo test -p connectrpc-codegen gate_client_feature`
- `cargo test -p connectrpc-build gate_client_feature`
- `cargo check -p connectrpc-codegen`
- `cargo check -p connectrpc-build`
- `cargo +nightly-2026-02-27 fmt --all -- --check`
- `RUSTDOCFLAGS="-Dwarnings" cargo doc --workspace --all-features --no-deps`
- `git diff --check`

## Risk

Low. This is additive, the default remains `client`, and the existing bare plugin flag continues to work. The main compatibility risk is rejecting an empty explicitly configured feature name; that is covered by a focused parser test.

## Out of scope

- JSON or message-type feature gating.
- Regenerating checked-in generated code; default output is unchanged.
- Replacing or changing the existing `gate_client_feature(bool)` API.

## Issue linkage

Closes #181

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,13 @@ increment the patch version.
 
 ### Added
 
+- **Configurable codegen client feature gate name** ([#181]).
+  `gate_client_feature=<name>` lets `protoc-gen-connect-rust` emit generated
+  client items behind `#[cfg(feature = "<name>")]` instead of the default
+  `client`; `connectrpc-build` exposes the same through
+  `Config::client_feature_name`. The existing bare `gate_client_feature`
+  option and `Config::gate_client_feature(true)` behavior continue to use
+  `client`.
 - **Optional `json` cargo feature for proto-only builds** ([#172]). The
   Connect JSON codec requires `serde::Serialize`/`Deserialize` on every
   message type, so the code generator derives them by default — pure cost for
--- a/connectrpc-build/src/lib.rs
+++ b/connectrpc-build/src/lib.rs
@@ -203,6 +203,19 @@ impl Config {
         self
     }
 
+    /// Set the Cargo feature name used by [`Config::gate_client_feature`]
+    /// (default: `"client"`).
+    ///
+    /// Use this when the generated crate exposes its client surface under a
+    /// different feature name, such as `grpc-client` or `transport`. This
+    /// method does not enable gating by itself; pair it with
+    /// `gate_client_feature(true)`.
+    #[must_use]
+    pub fn client_feature_name(mut self, feature: impl Into<String>) -> Self {
+        self.options.client_feature_name = feature.into();
+        self
+    }
+
     /// Replace the underlying buffa [`CodeGenConfig`] wholesale.
     ///
     /// Any buffa knob not surfaced as a builder method here can be set this
@@ -707,6 +720,7 @@ mod tests {
             .generate_json(false)
             .emit_register_fn(false)
             .gate_client_feature(true)
+            .client_feature_name("grpc-client")
             .include_file("_inc.rs");
         assert_eq!(cfg.files.len(), 2);
         assert_eq!(cfg.includes.len(), 1);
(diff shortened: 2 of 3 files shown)


iainmcgin (approved):
[claude code] Thanks — this is a clean implementation of #181 and the default behaviour is fully backward-compatible (`task generate:all` produces zero diff in the checked-in `*.__connect.rs` files).

I pushed two maintainer commits:

- **92d384b** — merged `main` and resolved the `CHANGELOG.md` conflict (keep-both; also added the missing `[#181]` link reference).
- **3390690** — three small tightenings:
  - **Validate the feature name against Cargo's grammar.** A name like `"grpc client"` or `"foo/bar"` would emit `#[cfg(feature = "...")]` that Cargo can never enable, silently suppressing every generated client item with no diagnostic. The check is inlined for now; once the workspace moves to `buffa-codegen` 0.8 it can defer to `buffa_codegen::FeatureGateNames::is_valid_name`.
  - **`Config::client_feature_name(..)` now also enables `gate_client_feature`**, mirroring the plugin's `gate_client_feature=<name>` form. Previously calling only `.client_feature_name("x")` was a silent no-op (gating stayed off). The order-dependence with an explicit `.gate_client_feature(false)` is documented on both setters.
  - Updated `gate_client_feature`'s rustdoc to cross-link the new setter instead of hard-coding `"client"`.

LGTM with those applied.

PR #196, opened 2026-06-24
Omit content-encoding for gzip values since the body is already decompressed by the browser

bickfordb:
I noticed that the WASM client was failing for responses that were compressed by intermediary proxies, and then decompressed automatically by the browser.  Wasn't sure if it makes sense to strip content-encoding entirely.

### DIFF
--- a/examples/wasm-client/src/transport.rs
+++ b/examples/wasm-client/src/transport.rs
@@ -141,6 +141,13 @@ async fn fetch(
             let Some(val) = pair.get(1).as_string() else {
                 continue;
             };
+            // If the content-encoding is gzip, we skip it because the browser automatically
+            // decompresses the response body, so we don't need to handle it ourselves. If we
+            // didn't skip it, the client would try to decompress an already decompressed body,
+            // which would result in an error.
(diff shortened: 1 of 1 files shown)


bickfordb:

iainmcgin (approved):
[claude code] Thanks — this is the right fix at the right layer. I pushed one follow-up commit (1972e89) to widen it slightly, since you'd already asked the right question in the description:

- Strip `Content-Encoding` **unconditionally** rather than only for `gzip`. Fetch transparently decodes `gzip`/`deflate`/`br`/`zstd`, and `Accept-Encoding` is a Fetch-forbidden request header that the browser replaces with its own list — so a server may legitimately respond with any of those regardless of what we advertised. With the `gzip`-only check, a `br` response would reach `CompressionRegistry::decompress_with_limit` and fail.
- Also strip `Content-Length`, which is equally stale relative to the decoded body (matches what `reqwest` does after auto-decompression).
- Swapped `to_lowercase()` for `eq_ignore_ascii_case` against the `http::header` constants (Fetch already lowercases iterator names, so this is just avoiding the per-header allocation).

LGTM with that applied.

PR #197, opened 2026-06-24
client: bound h2 connection establishment by default; proxy keep-alive knobs

rpb-ant:
## What

Makes `Http2Connection` (and `HttpClient`) bound connection establishment by default, so a hung or half-open server cannot stall `poll_ready` indefinitely. Surfaces the common HTTP/2 settings (keep-alive, window sizes) directly on the new builder so callers don't have to construct a `hyper::client::conn::http2::Builder` by hand.

This carries the first commit from #187 and reshapes the API in four further commits. It is an alternative to #187; one of the two should land.

## Why

Issue #137: the `Reconnect` state machine's `Connecting` future runs DNS → TCP → TLS → h2 preface serially with no time bound on any step. A server that accepts the TCP connection but never completes the TLS handshake (e.g. a draining pod whose kernel backlog still accepts) parks that future forever, which parks `poll_ready` for every caller sharing the connection. h2 keep-alive can't help because the h2 layer doesn't exist yet.

#187 added the timeout knobs but left them opt-in (default unbounded), left DNS uncovered by either knob, and gave callers who set h2 keep-alive (the `with_builder_tls` path) no way to reach them.

## API

`Http2ConnectionBuilder` (via `Http2Connection::builder()`) is now the single config surface for every transport — plaintext, TLS, caller-supplied connector, Unix socket. Every existing `Http2Connection` constructor delegates through it.

- `establishment_timeout(Duration)` — wall-clock bound on DNS + TCP + TLS + h2 preface as one budget. **Defaults to `DEFAULT_ESTABLISHMENT_TIMEOUT` (20s)**, matching grpc-go's `MinConnectTimeout`. Exceeding it surfaces as `ErrorCode::Unavailable`. Pass `Duration::MAX` to opt out.
- `tcp_connect_timeout(Duration)` — per-address TCP `connect(2)` budget on the built-in connector, divided across the resolved address set, so the connector moves on to the next address rather than burning the whole establishment budget on one blackholed peer. **Defaults to `DEFAULT_TCP_CONNECT_TIMEOUT` (5s).** Ignored on the custom-connector / Unix-socket terminals (no built-in connector there).
- `keep_alive_interval` / `keep_alive_timeout` / `keep_alive_while_idle` / `initial_stream_window_size` / `initial_connection_window_size` / `adaptive_window` — proxied hyper h2 setters, matching `tonic::Endpoint`. A `TokioTimer` is pre-wired so the keep-alive setters work without hyper's "supply a timer" panic.
- `h2_settings(Builder)` — escape hatch for hyper knobs not proxied above; replaces prior h2 settings wholesale.
- Terminals: `lazy_plaintext` / `connect_plaintext` / `lazy_tls` / `connect_tls` / `lazy_with_connector` / `connect_with_connector` / `lazy_unix` / `connect_unix`.

`HttpClientBuilder` gains `establishment_timeout(Duration)` (DNS + TCP + TLS via a `TimeoutConnector` wrapper; the h2 preface runs inside hyper's pool and is not separately observable). `connect_timeout` keeps its shipped name. Both default on with the same constants.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,23 @@ increment the patch version.
 
 ### Added
 
+- **`Http2ConnectionBuilder` with establishment-timeout and HTTP/2 keep-alive
+  knobs** ([#137], [#197]). `Http2Connection::builder()` is now the single
+  configuration surface for every transport flavour (plaintext, TLS,
+  caller-supplied connector, Unix socket); the existing `Http2Connection`
+  constructors are shortcuts for it with default settings. The builder adds
+  `establishment_timeout` (a wall-clock bound on DNS + TCP + TLS + HTTP/2
+  preface as one budget) and `tcp_connect_timeout` (a per-address TCP bound on
+  the built-in connector), and proxies hyper's `keep_alive_interval`,
+  `keep_alive_timeout`, `keep_alive_while_idle`, `initial_stream_window_size`,
+  `initial_connection_window_size`, and `adaptive_window` setters directly,
+  with a `TokioTimer` pre-wired so the keep-alive setters work without hyper's
+  "supply a timer" panic. `h2_settings(|b| ...)` is the escape hatch for hyper
+  knobs not proxied. `HttpClientBuilder` gains the matching
+  `establishment_timeout` (covering DNS + TCP + TLS) and `tcp_connect_timeout`
+  (the canonical name for [#117]'s `connect_timeout`, which remains as an
+  alias). Both builders also expose `no_establishment_timeout` /
+  `no_tcp_connect_timeout` to opt out of the new defaults below.
 - **Optional `json` cargo feature for proto-only builds** ([#172]). The
   Connect JSON codec requires `serde::Serialize`/`Deserialize` on every
   message type, so the code generator derives them by default — pure cost for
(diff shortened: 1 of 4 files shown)


iainmcgin:
[claude code] Reviewed alongside #187 — this is the better shape of the two and #187 should close in favour of it. The wiring is correct: `establishment_timeout` reaches all four transport paths via `run_establishment`, `tcp_connect_timeout` reaches `HttpConnector::set_connect_timeout` on the two paths where it applies, and `TimeoutConnector` satisfies hyper-util's sealed `Connect` bound by preserving `C::Response` (so `Connected` metadata survives). No correctness issues.

Pushed `4f66541` with the review fixups so we don't round-trip on the doc/API-polish items:

**API**
- `h2_settings` now takes `impl FnOnce(&mut Builder<TokioExecutor>)` and mutates in place — composes with the proxied keep-alive/window setters in either order instead of replacing them wholesale (the previous shape silently discarded earlier setters).
- `no_tcp_connect_timeout()` / `no_establishment_timeout()` on both builders as the explicit opt-out, replacing the doc-only `Duration::MAX` sentinel (which still works via `finite()`). The sentinel was inconsistent — `keep_alive_*(MAX)` doesn't honour it.
- `HttpClientBuilder::tcp_connect_timeout` is the canonical name, matching `Http2ConnectionBuilder`; `connect_timeout` stays as an alias.
- `with_builder_plaintext` / `with_builder_tls` deprecated toward `builder()` (the names now collide with the new entry point).
- Struct-level `#[must_use]` on both builders; the per-method attributes on `-> Self` setters dropped per `clippy::double_must_use`.

**Docs / CHANGELOG**
- Every shorthand constructor now states the default establishment bounds and points at `builder()` — the existing rustdoc didn't surface that they'd become bounded.
- CHANGELOG `[Unreleased]` Added/Changed/Deprecated entries; the 0.7.0 #117 entry gets a forward-reference note since its "behaviour is unchanged for callers who don't opt in" line is now superseded. The Changed entry calls out the per-address division of the 5s TCP budget.
- guide.md gains a "Connection-establishment bounds and keep-alive" section.
- `Duration::ZERO` documented as fail-fast on the timeout setters.

**Tests / internals**
- All three TEST-NET-1 tests hardened against transparent-proxy hosts (the two `HttpClient` ones could otherwise *hang*, since the HTTP/1.1 request has no timeout once the connect succeeds); the RFC 5737 comment softened to what the RFC actually says.
- `establishment_timeout_applies_with_custom_h2_settings` now also drives `.h2_settings(|b| ...)` so the closure path is exercised.
- Dropped the unused `E: 'static` bound on `run_establishment`.

Left as-is for now (your call): `connect_timeout` is a soft alias rather than `#[deprecated]`, since it shipped in 0.7.0 and a one-release grace seems kinder; happy to flip if you'd rather steer harder.

iainmcgin (approved):
[claude code] LGTM — wiring verified correct across all four transport paths, defaults match grpc-go's MinConnectTimeout, and the docs/CHANGELOG now make the default-on flip explicit. Supersedes #187 (closed).

PR #198, opened 2026-06-27
Reject malformed Connect END_STREAM JSON

fallintoplace:
## Summary
- reject malformed Connect END_STREAM JSON instead of defaulting to a clean success
- use one strict END_STREAM parser for both server-streaming and client-streaming responses
- add regression coverage for malformed END_STREAM frames after a valid message

## Why
Malformed END_STREAM payloads were being parsed with `unwrap_or_default()`, which turned invalid JSON into an empty terminal response and let corrupted or truncated stream terminators look successful.

## Impact
- malformed Connect END_STREAM frames now surface as terminal internal errors
- client-streaming keeps response headers attached on the new error path

## Validation
- `cargo test -p connectrpc malformed_end_stream`
- `cargo test -p connectrpc end_stream`

### DIFF
--- a/connectrpc/src/client/mod.rs
+++ b/connectrpc/src/client/mod.rs
@@ -2594,8 +2594,10 @@ where
             Err(e) => return e.into(),
         };
 
-        let end_stream: ClientEndStreamResponse =
-            serde_json::from_slice(&end_stream_data).unwrap_or_default();
+        let end_stream = match parse_connect_end_stream(&end_stream_data) {
+            Ok(end_stream) => end_stream,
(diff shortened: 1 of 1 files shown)


iainmcgin:
[claude code] Closing this as a duplicate: the patch here is byte-for-byte identical to what already landed on `main` as #192 (commit dcc7830, "Reject malformed Connect END_STREAM JSON"). That's also why this PR shows as conflicting — the same hunks are already present on `main`.

Thanks for the fix — it's in! The remaining open PRs (#199, #200, #201) are being reviewed now.

PR #199, opened 2026-06-27
Preserve transport ConnectErrors in client calls

fallintoplace:
## Summary
- preserve transport-produced `ConnectError`s instead of remapping them to `Unavailable`
- apply the shared transport error mapper across unary, unary GET, server-streaming, client-streaming, and bidi client calls
- add regressions covering the real plaintext-client-to-`https://` configuration mismatch across all five call shapes

## Why
`HttpClient::send()` already classifies local scheme/config mismatches as `InvalidArgument`, but the higher-level client call helpers wrapped every transport error as `Unavailable`. That made local caller bugs look like retryable outages and obscured the real cause.

## Impact
- local transport/configuration mistakes now retain their original `ConnectError` classification
- generic transport failures still map to `Unavailable`
- wrapped transport errors that carry a `ConnectError` in their source chain are preserved too

## Validation
- `cargo test -p connectrpc --features client preserves_transport_connect_error`
- `cargo test -p connectrpc transport_send_error_mapper_preserves_connect_error_in_source_chain`
- `cargo test -p connectrpc --features client http_client_plaintext_rejects_https`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,19 @@ increment the patch version.
 
 ### Changed
 
+- **Client transport errors keep their original classification** ([#199]).
+  The client call paths previously wrapped every transport `send` failure as
+  `unavailable` with a `request failed:` prefix, including errors that were
+  already classified `ConnectError`s — so a local configuration mistake such
+  as pointing `HttpClient::plaintext()` at an `https://` URL looked like a
+  retryable outage. A `ConnectError` found anywhere in the transport error's
+  source chain is now surfaced verbatim (code, message, details, and attached
+  metadata; the `request failed:` prefix is gone for the built-in
+  transports). Transport errors with no `ConnectError` in their chain are
+  wrapped as `unavailable`, unchanged. Retry classifiers keyed on
+  `unavailable` keep matching genuine transport outages, and now correctly
+  stop retrying non-retryable local errors; anything matching on the
+  `request failed:` message prefix should match on the error code instead.
 - **Client connection establishment is bounded by default** ([#137], [#197]).
   `Http2Connection` and `HttpClient` now bound connection establishment to
   `DEFAULT_ESTABLISHMENT_TIMEOUT` (20s, matching grpc-go's `MinConnectTimeout`)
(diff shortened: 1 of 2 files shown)


iainmcgin (approved):
[claude code] Reviewed with two parallel review passes (correctness + API ergonomics), findings verified against the worktree. The change is correct and well-aimed: both built-in transports already produce classified `ConnectError`s (`HttpClient` and `SharedHttp2Connection` both have `Error = ConnectError`), so the old wrap was destroying the code, details, and attached metadata of an already-classified error — and made non-retryable local mistakes look like retryable outages. The five per-path tests all carry a code assertion that fails on the pre-PR code, including the non-obvious bidi deferred-error plumbing through `response_task` → `construct_err` → both `message()` and `error()`. Good tests.

One reviewer finding worth recording because it was *refuted*: the concern that a `ConnectError` boxed in `Box<dyn Error + Send + Sync>` (tower middleware convention) would defeat the chain walk. It can't — `Box<dyn Error>` does not itself implement `std::error::Error` (the blanket impl requires a sized type), so a boxed trait object can neither satisfy the `ClientTransport::Error` bound nor appear as a link in a `source()` chain. The walk as written is exhaustive. I added a comment on the helper documenting this so it doesn't get "fixed" later.

I pushed a fixup commit to the branch (rebased onto current main first):

- **CHANGELOG entry** under `### Changed` — this is an observable error-classification change (transport errors are no longer uniformly `unavailable`, and the `request failed:` message prefix is gone for the built-in transports), and consumers with retry classifiers or log matchers need the migration note.
- **Doc comments** on the two new helpers stating the policy (preserve code/message/details/metadata; drop outer wrappers' `Display` context), and a contract paragraph on `ClientTransport::Error` — the chain walk turns the transport error type into a deliberate side channel ("return a `ConnectError` anywhere in your source chain to control the surfaced classification"), which custom-transport authors need to know exists.
- Renamed the helper's `message` parameter to `context` (it's a prefix, not the message).

Known remaining inconsistency, deferred: mid-stream body-read errors are still flattened to `internal` (three sites). Different phase, needs a body-error bound change; tracked in the follow-up issue on error metadata consistency. Approving.

PR #200, opened 2026-06-27
Tighten gRPC response content-type validation

fallintoplace:
## Summary
- replace the gRPC unary response `content-type` prefix check with strict protocol-and-codec-aware validation
- allow the bare proto defaults (`application/grpc` and `application/grpc-web`) while still stripping parameters
- add regressions for rejecting gRPC-Web on a gRPC client, rejecting mismatched response codecs, and accepting `application/grpc; charset=utf-8`

## Why
The previous `starts_with("application/grpc")` check accepted `application/grpc-web*` and never verified that the response codec matched the client configuration. That let a gRPC client accept gRPC-Web framing and could let a proto-configured client try to decode JSON bytes as proto.

## Impact
- gRPC unary responses now require an exact allowed content type for the configured protocol and codec
- `application/grpc` remains accepted as the default proto subtype, even with parameters
- gRPC-family mismatches are rejected before body parsing starts

## Validation
- `cargo test -p connectrpc grpc_unary_`
- `cargo test -p connectrpc grpc_web_unary_stops_reading_after_trailer_frame`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -164,6 +164,19 @@ increment the patch version.
 
 ### Fixed
 
+- **gRPC unary response content-type validation is protocol- and
+  codec-aware** ([#200]). The client previously accepted any response
+  `content-type` starting with `application/grpc`, so a gRPC client could
+  accept gRPC-Web framing, and a proto-configured client could try to decode
+  JSON bytes as proto. Validation now mirrors connect-go's
+  `grpcValidateResponseContentType`: the exact configured subtype and the
+  bare family type (`application/grpc` / `application/grpc-web`, which imply
+  proto and are what proxies such as Envoy send on trailers-only error
+  replies) are accepted, with `; parameter` suffixes stripped; a same-family
+  codec mismatch is rejected as `internal` and anything else as `unknown`,
+  with the expected content type named in the error message. A missing
+  `content-type` header remains accepted. This also covers gRPC / gRPC-Web
+  client-streaming responses, which share the unary parse path.
 - **Malformed Connect END_STREAM JSON now returns `internal`** ([#192],
   [#201]). A streaming Connect response whose END_STREAM envelope body was not
   valid JSON was silently treated as a clean `Ok(None)` close (the parse error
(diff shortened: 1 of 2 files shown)


iainmcgin (dismissed):
[claude code] Reviewed with two parallel review passes (correctness + API ergonomics), findings verified against the worktree and against connect-go's `grpcValidateResponseContentType`. The direction is exactly right — the old `starts_with("application/grpc")` check let a gRPC client accept gRPC-Web framing and a proto client try to decode JSON — and the structure (parameter stripping, bare-type acceptance, headers attached to the rejection) was all correct. The client-streaming gRPC path also gains the validation for free via the shared parse path, which is a bonus.

I pushed a fixup commit (after rebasing onto current main) that aligns the edge cases with connect-go and closes the review findings:

- **Bare family types are now accepted for *any* codec, not just proto.** Both reviewers converged on this independently: proxies that synthesize trailers-only error replies (Envoy local replies being the canonical case) send bare `application/grpc` regardless of the request subtype, so a `+json` client would have turned a retryable `unavailable`-class failure into a dead-end `internal`. connect-go accepts the bare type whenever the request carries a `+codec` subtype — which this client always does.
- **Cross-family mismatches are now `unknown`, not `internal`**, matching connect-go's classification: `internal` is reserved for same-family `+`-subtype mismatches ("a broken server/intermediary produced this"), `unknown` for everything that doesn't look like a response in the configured family at all. The `grpc_unary_rejects_grpc_web_content_type` test assertion is updated accordingly.
- **The error message now names the expected type** (`unexpected content-type: X (expected Y)`) — the highest-leverage diagnostic improvement; the gRPC-client-pointed-at-gRPC-Web misconfiguration is now self-diagnosing instead of a source dive.
- **Coverage for the previously untested branches**: `unknown` rejection, bare `application/grpc-web` acceptance, missing content-type acceptance, bare-with-JSON-codec acceptance, and the gRPC-Web-client-receives-gRPC direction — as direct unit tests on the helper.
- **Doc comment on the helper** recording the policy and the connect-go reference, a `split_once` normalization (the `split().next().unwrap_or()` fallback was unreachable), and a **CHANGELOG entry** under `### Fixed` since this newly rejects responses that previously parsed.

Known gap, deferred with a follow-up issue: the streaming/bidi gRPC paths (`make_server_stream`) still have no content-type validation at all, so the exact misroute this PR catches on unary still fails as envelope-parse garbage on a stream. The helper is shaped to be reusable there. Approving.

iainmcgin (approved):

PR #221, opened 2026-07-07
client: make BidiStream receive initialization cancellation-safe

EffortlessSteven:
## What this does

Cancelling the initial `BidiStream::message()` could leave `RecvState` failed without a real error while response initialization was still pending. Later reads returned `internal("stream in failed state")` even when the server completed normally.

This is the `BidiStream::message()` cancellation-safety follow-up called out in connectrpc/connect-rust#159.

**Fix:** keep response initialization in `RecvState` while awaiting response headers and `ServerStream` construction. Dropping the `message()` future leaves the active stage resumable; real transport, deadline, and protocol failures become sticky `Failed(ConnectError)` states.

## Verification

| Check | Result |
| --- | --- |
| initialization cancellation | before headers and during Connect error-body parsing: unpatched `stream in failed state`, patched resumes or returns the server error |
| sticky failures | deadline before headers, deadline during construction, and transport failure stay sticky |
| preservation coverage | auto traits and half-duplex send-before-read behavior preserved |
| local package gates | `connectrpc` all-features, no-default-features, clippy, fmt, docs, and workspace check clean |

## Review map

- `RecvState`: replace `Pending` and `Failed` + `construct_err` with `AwaitingHeaders`, `Constructing`, `Ready`, and `Failed(ConnectError)`. Pending initialization now stays in the state while awaited, and terminal initialization failures carry their actual error.
- `BidiStream::message`: poll retained header and construction handles in place under the call deadline, then transition only after completion to `Ready` or `Failed(error)`.
- `call_bidi_stream`: still spawns `transport.send(...)` immediately, preserving half-duplex send-before-read behavior.
- `error()` / `Debug`: read terminal initialization failures from `RecvState::Failed`.

### DIFF
--- a/.changes/unreleased/Fixed-20260706-203937.yaml
+++ b/.changes/unreleased/Fixed-20260706-203937.yaml
@@ -1,0 +1,8 @@
+kind: Fixed
+body: |-
+  **Initial `BidiStream::message()` cancellation no longer breaks the receive side**.
+  Dropping the first receive while it is waiting for response headers or Connect
+  error-body parsing now leaves initialization pending for the next `message()`
+  call, while real transport, deadline, and protocol failures stay terminal and
+  sticky.
+time: 2026-07-06T20:39:37.361463737-04:00
--- a/.changes/unreleased/Fixed-20260706-203937.yaml
+++ b/.changes/unreleased/Fixed-20260706-203937.yaml
@@ -4,5 +4,12 @@ body: |-
   Dropping the first receive while it is waiting for response headers or Connect
   error-body parsing now leaves initialization pending for the next `message()`
   call, while real transport, deadline, and protocol failures stay terminal and
-  sticky.
+  sticky. Dropping the `BidiStream` (or failing the call at its deadline) now
+  aborts the in-flight initialization task instead of detaching it: a stalled
+  server can no longer pin an abandoned call's resources, and dropping the
+  stream cancels the call outright rather than flushing buffered request
+  messages in the background — callers that need the request delivered must
+  drive the call via `message()` before dropping. `message()` now requires
+  `B: 'static` and the view type to be `'static`, which every transport and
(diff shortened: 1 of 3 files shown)


iainmcgin:
[claude code] Thanks — the state machine is sound and the tests are well constructed. I verified the fix end-to-end (cancel/resume against a real server, deadline stickiness against a stalling server, plus a before/after against the pre-fix code) and everything holds.

I pushed one maintainer commit (cbfb34e) on top rather than round-tripping review comments, addressing two things the review surfaced:

- **Task lifetime on abandonment**: a task left in `AwaitingHeaders`/`Constructing` detached on drop, so with no call deadline a server that stalls response HEADERS or a Connect error body could pin the task and its connection indefinitely. `BidiStream` now aborts the in-flight task in `Drop`, and the deadline-failure arm aborts the headers task instead of detaching it. This is documented as drop-cancels-the-call semantics (matching connect-go/tonic convention): buffered-but-untransmitted request messages are not flushed in the background on drop.
- **Smaller items**: changelog note for the new `B: 'static`/view `'static` bounds on `message()` (already implied at the only construction site, so no practical break), join-failure messages reworded since `JoinError` covers cancellation as well as panic, `headers()`/`error()` docs now describe the retained-initialization window, single-match `Debug` impl, and two stale comments.

Regression tests for the three new abort paths are included. One adjacent finding — `call_client_stream` has the same detach-on-cancel leak this PR fixes for `BidiStream` — is pre-existing on `main` and will be filed as a follow-up issue rather than expanding this PR.

iainmcgin (approved):
[claude code] Reviewed in depth (state-machine trace, adversarial verifier pass, end-to-end cancel/resume + deadline drive against a real server, 20/20 stress runs). Sound fix; maintainer follow-ups applied in cbfb34e. Approving and merging.

PR #226, opened 2026-07-09
client: stop the client-stream transport task when the call is abandoned

### LINKED ISSUE #224: client: call_client_stream detaches its transport-send task — same leak class as #221
## Summary

`call_client_stream` has the same detach-on-cancel task leak that #221 fixed for `BidiStream`: the transport send runs in a detached task whose lifetime nothing bounds once the caller stops waiting.

## Details

In `connectrpc/src/client/mod.rs`, `call_client_stream` spawns the transport send and discards the handle:

```rust
let response_fut = transport.send(http_request);
let (resp_tx, resp_rx) = tokio::sync::oneshot::channel::<...>();
let _ = crate::spawn_detached(async move {
    let result = response_fut.await.map_err(...);
    let _ = resp_tx.send(result);
});
```

Two paths orphan the task:

- **Deadline expiry**: the `with_deadline` wrapper bounds only the caller's wait on `resp_rx`; on expiry the awaiting future is dropped and the detached task keeps running. Note the deadline does not bound the task itself — the leak is unbounded whether or not a deadline was set; the deadline only converts the caller's hang into an error return.
- **Caller cancellation**: dropping the `call_client_stream` future drops `resp_rx`, but the spawned task's code is linear (`response_fut.await` then `resp_tx.send`) — it only observes the closed receiver *after* `transport.send` resolves. It never selects against `resp_tx.closed()`.

Dropping `tx` ends the `ChannelBody` cleanly (END_STREAM, not a reset), so the request completes — and then a server that stalls response HEADERS without ending the HTTP/2 stream pins the detached task, its stream, and the transport/connection reference indefinitely.

## Suggested fix

Mirror the #221 pattern: keep the `JoinHandle` and abort it on deadline failure and on caller cancellation. Since `call_client_stream` is a plain `async fn`, cancellation needs either a drop-guard around the handle or a `select!` against `resp_tx.closed()` inside the spawned task.

A regression test can reuse the shape of the #221 tests: a response task gated on a never-resolving future holding a guard oneshot sender, asserting the guard drops (task aborted) after deadline expiry / caller drop.

EffortlessSteven:
## What this does

Abandoning a `call_client_stream` call (dropping its future or hitting the deadline) while the transport is still awaiting response headers leaves the background `transport.send(...)` running indefinitely: the detached task driving it only checks the receiver after the send resolves.

**Fix:** the spawned task now stops the send when the caller stops waiting (future dropped or deadline fired). The response and transport-error paths are unchanged.

Note that this is also a behavior change, not only a leak fix: abandonment now cancels the underlying request, so a request that was still being sent may never reach the server. The `# Cancellation` docs and the changelog entry call this out, matching the equivalent `BidiStream` documentation from #221.

## Verification

Unpatched, the abandonment tests time out: the transport send future is never dropped.

| Check | Result |
| --- | --- |
| abandon: deadline / cancel / mid-drain | unpatched: send never dropped; patched: dropped |
| success, transport-error paths | unchanged |
| `fmt`, `clippy -D warnings`, `test` | clean / 562 passed |

## Review map

- `call_client_stream`: the spawned send task selects `transport.send(...)` against `resp_tx.closed()`, so a dropped receiver (cancellation or deadline) stops the send instead of leaving it detached.
- `connectrpc/Cargo.toml`: the always-compiled `call_client_stream` now uses `tokio::select!`, so the base tokio dep gains `macros`; the `server` feature's own `tokio/macros` entry is dropped as redundant, with a note on the base dependency explaining why `macros` is load-bearing.
- tests: pin both abandonment paths dropping the send future (with mid-drain backpressure), and preserve the success and transport-error paths.
- maintainer fixup (`6f67559`): documents the delivery caveat in the rustdoc and changelog, adds a `# Cancellation` note to `call_server_stream`, and consolidates the `tokio/macros` feature declaration.

Fixes #224

### DIFF
--- a/.changes/unreleased/Fixed-20260709-051015.yaml
+++ b/.changes/unreleased/Fixed-20260709-051015.yaml
@@ -3,7 +3,10 @@ body: |-
   **`call_client_stream` now stops its background transport send when the call
   is abandoned** ([#224]). Dropping the call future or hitting its deadline
   while the transport is still waiting for response headers no longer leaves the
-  detached send task polling the transport indefinitely.
+  detached send task polling the transport indefinitely. Note that abandonment
+  now cancels the underlying request: a request that was still being sent when
+  the call was abandoned may never reach the server, so a caller that needs
+  the request delivered must drive the call to completion.
 
   [#224]: https://github.com/connectrpc/connect-rust/issues/224
 time: 2026-07-09T05:10:15.451894063-04:00
--- a/.changes/unreleased/Fixed-20260709-051015.yaml
+++ b/.changes/unreleased/Fixed-20260709-051015.yaml
@@ -1,0 +1,9 @@
+kind: Fixed
+body: |-
+  **`call_client_stream` now stops its background transport send when the call
+  is abandoned** ([#224]). Dropping the call future or hitting its deadline
+  while the transport is still waiting for response headers no longer leaves the
+  detached send task polling the transport indefinitely.
+
+  [#224]: https://github.com/connectrpc/connect-rust/issues/224
+time: 2026-07-09T05:10:15.451894063-04:00
--- a/connectrpc/Cargo.toml
+++ b/connectrpc/Cargo.toml
@@ -24,7 +24,10 @@ bytes = { workspace = true }
 # Tower service abstraction (the foundation)
 tower = { workspace = true }
 
-# Async runtime (minimal features for core)
+# Async runtime (minimal features for core). `macros` is load-bearing for
+# `tokio::select!` in both the always-compiled client-stream path and the
+# `server` accept loop; CI's `--all-targets` would hide its removal via
+# dev-dep unification while breaking downstream `server`-only builds.
 tokio = { workspace = true, features = ["rt", "io-util", "sync", "time", "macros"] }
 futures = { workspace = true }
 pin-project = { workspace = true }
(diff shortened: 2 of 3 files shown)


iainmcgin (approved):
[claude code] Reviewed on Iain's behalf; approving.

Verification performed locally:

- Reverted just the fix (keeping the new tests) and confirmed all three abandonment tests fail unpatched and pass patched — they genuinely gate the bug.
- Confirmed the ownership invariant the fix relies on: `resp_rx` lives inside the `timeout_at`-wrapped block, so both abandonment paths (future drop, deadline expiry) drop it, and the spawned task's `closed()` branch then drops the in-flight send. The simultaneous-readiness race is benign in both directions.
- Confirmed this was the last detached transport-send spawn in the client (`call_bidi_stream` keeps its `JoinHandle` per #221; `call_server_stream` awaits the send inline), closing out the #224 leak class.

I pushed a small maintainer fixup (6f67559) rather than round-tripping review comments: it documents the caller-visible consequence (an abandoned call now cancels the underlying request, so a request still being sent may never reach the server — the same caveat #221 documented for `BidiStream`), adds the matching `# Cancellation` note to `call_server_stream`, consolidates the now-redundant `server`-feature `tokio/macros` entry into a note on the base dependency, and rewords a few test comments into code-present tense.

Thanks for the contribution — the `started_rx` gate proving the drop is of an in-flight (not merely unpolled) future is a nice touch, and the mid-drain backpressure case is exactly the coverage this needed.

PR #227, opened 2026-07-09
client: take an async Stream for client-streaming requests

iainmcgin:
## Summary

Client-streaming calls now take an async `Stream` of requests instead of a synchronous iterator.

`call_client_stream` and the generated client-streaming methods accept `impl ClientRequestStream<Req>` — a sealed trait implemented automatically for every `Stream<Item = Req> + Send + 'static` — so request messages can be produced as they become available — paced by timers, read from sockets, forwarded from channels — without buffering the whole upload or blocking a thread. The synchronous-iterator shape could not express this: there was no way to await between messages, which ruled out pump-style uploads (e.g. forwarding chunks under a per-chunk idle deadline).

**Breaking change**, targeted at 0.9.0. Migration for a ready collection is one line, with no new dependency:

```rust
// before
client.sum(requests).await?;
// after
client.sum(connectrpc::client::stream_iter(requests)).await?;
```

`Stream` (the `futures` 0.3 trait) and `stream_iter` (`futures::stream::iter`) are re-exported from `connectrpc::client`, and every generated client-streaming method carries the migration hint in its rustdoc. The bound trait carries `#[diagnostic::on_unimplemented]`, so passing a collection directly produces compiler output that names `stream_iter` and explains the `Send + 'static` requirement, rather than a bare unmet-trait error.

## Design: the stream is the request body

The stream is not pumped by a library-side drain loop — it backs the HTTP request body directly (`EncodingBody`, which envelope-encodes each message in `poll_frame`). The transport polls for the next frame only while it can send, so:

- backpressure is HTTP/2 flow control (peak memory ~one envelope, previously a 32-envelope channel);
- a server that ends the RPC mid-upload produces a response and the call returns, even while the caller's stream is idle — previously unobservable until the deadline;
- a server that sends response headers early while continuing to read the upload keeps receiving messages (a drain loop racing the response future would truncate here — caught by review and pinned with a regression test).

An encode failure is emitted through the body (aborting the request) and mirrored into the call's return value so the caller sees the precise error. The stream is held in a `sync_wrapper::SyncWrapper` (new tiny zero-dep dependency, the axum pattern) so the boxed body is `Sync` without requiring `Sync` of the caller's stream. The previous channel + detached-send-task machinery for client-streaming is deleted; bidi is unchanged (its push-style `send()` API keeps the channel).

## Changes

- `connectrpc/src/client/mod.rs`: signature, `EncodingBody`, `Stream`/`stream_iter` re-exports, rewritten docs with `# Errors`.
- `connectrpc-codegen/src/codegen.rs`: generated arg type + doc hint; checked-in generated dirs regenerated.
- Call sites updated: conformance client, benches, streaming-tour example, e2e tests; `docs/guide.md` shows both `stream_iter` and a channel-backed producer.
- New e2e tests: async producer paced by a channel; early-response-headers-do-not-truncate-upload; early-server-response-unblocks-pending-stream.

## Testing

- `task lint` / `task test` green; clippy all-targets `-D warnings` clean; docs with broken-links denied; MSRV (1.88), wasm32, and minimal-features checks pass.
- Conformance client suites: Connect 2580, gRPC 1454, gRPC-Web 2838 — all passing.
- Live socket drives (HTTP/1.1 and HTTP/2): a producer pacing 3 messages 100 ms apart aggregates correctly (elapsed ≥ 300 ms proves incremental sending); a handler rejecting without draining returns in ~2 ms against a never-yielding stream (previously an unbounded hang).
- `cargo semver-checks` vs 0.8.1 passes (the break is in an `impl Trait` argument bound, outside its lint coverage); declared in the changelog fragment.

### DIFF
--- a/connectrpc-codegen/src/codegen.rs
+++ b/connectrpc-codegen/src/codegen.rs
@@ -2254,7 +2254,13 @@ fn generate_client_method(
             #[doc = " are sent as the stream yields them. It backs the request"]
             #[doc = " body, so yield owned messages or feed the call from a"]
             #[doc = " channel-backed stream. For a collection that is already in"]
-            #[doc = " hand, wrap it with `::connectrpc::client::stream_iter(...)`."]
+            #[doc = " hand, wrap it with `::connectrpc::stream_iter(...)`."]
+            #[doc = ""]
+            #[doc = " Dropping the returned future cancels the call: the request"]
+            #[doc = " body is dropped along with it, so messages the stream had"]
+            #[doc = " not yet yielded are never delivered. A caller that needs the"]
+            #[doc = " request delivered must drive the call to completion rather"]
+            #[doc = " than, say, wrapping it in a `timeout`."]
         };
         ret_ty = quote! {
             Result<
--- a/connectrpc-codegen/src/codegen.rs
+++ b/connectrpc-codegen/src/codegen.rs
@@ -2249,9 +2249,10 @@ fn generate_client_method(
         // Client-stream
         extra_doc = quote! {
             #[doc = ""]
-            #[doc = " `requests` is an asynchronous stream; messages are sent as the"]
-            #[doc = " stream yields them. It must be `Send + 'static` (it backs the"]
-            #[doc = " request body), so yield owned messages or feed the call from a"]
+            #[doc = " `requests` is any `Stream<Item = ...> + Send + 'static` of"]
+            #[doc = " request messages (the `ClientRequestStream` bound); messages"]
+            #[doc = " are sent as the stream yields them. It backs the request"]
+            #[doc = " body, so yield owned messages or feed the call from a"]
             #[doc = " channel-backed stream. For a collection that is already in"]
             #[doc = " hand, wrap it with `::connectrpc::client::stream_iter(...)`."]
         };
@@ -2268,8 +2269,9 @@ fn generate_client_method(
                 requests, options,
             ).await
         };
-        short_args = quote! { requests: impl ::connectrpc::client::Stream<Item = #input_type> + Send + 'static };
-        opts_args = quote! { requests: impl ::connectrpc::client::Stream<Item = #input_type> + Send + 'static, options: ::connectrpc::client::CallOptions };
+        short_args =
+            quote! { requests: impl ::connectrpc::client::ClientRequestStream<#input_type> };
+        opts_args = quote! { requests: impl ::connectrpc::client::ClientRequestStream<#input_type>, options: ::connectrpc::client::CallOptions };
         short_delegate_args = quote! { requests, ::connectrpc::client::CallOptions::default() };
     } else if client_streaming && server_streaming {
         // Bidi
--- a/connectrpc-codegen/src/codegen.rs
+++ b/connectrpc-codegen/src/codegen.rs
@@ -2242,9 +2242,19 @@ fn generate_client_method(
     let short_args: TokenStream; // args to the no-opts convenience method
     let opts_args: TokenStream; // args to the _with_options method
     let short_delegate_args: TokenStream; // how short delegates to opts
+    // Extra doc lines appended to both methods (client-stream input contract).
+    let mut extra_doc = quote! {};
 
     if client_streaming && !server_streaming {
         // Client-stream
+        extra_doc = quote! {
+            #[doc = ""]
+            #[doc = " `requests` is an asynchronous stream; messages are sent as the"]
+            #[doc = " stream yields them. It must be `Send + 'static` (it backs the"]
+            #[doc = " request body), so yield owned messages or feed the call from a"]
+            #[doc = " channel-backed stream. For a collection that is already in"]
+            #[doc = " hand, wrap it with `::connectrpc::client::stream_iter(...)`."]
+        };
         ret_ty = quote! {
             Result<
                 ::connectrpc::client::UnaryResponse<::buffa::view::OwnedView<#output_view_type<'static>>>,
@@ -2258,8 +2268,8 @@ fn generate_client_method(
                 requests, options,
             ).await
         };
-        short_args = quote! { requests: impl IntoIterator<Item = #input_type> };
-        opts_args = quote! { requests: impl IntoIterator<Item = #input_type>, options: ::connectrpc::client::CallOptions };
+        short_args = quote! { requests: impl ::connectrpc::client::Stream<Item = #input_type> + Send + 'static };
+        opts_args = quote! { requests: impl ::connectrpc::client::Stream<Item = #input_type> + Send + 'static, options: ::connectrpc::client::CallOptions };
         short_delegate_args = quote! { requests, ::connectrpc::client::CallOptions::default() };
     } else if client_streaming && server_streaming {
         // Bidi
@@ -2320,11 +2330,13 @@ fn generate_client_method(
 
     Ok(quote! {
         #[doc = #doc]
+        #extra_doc
         pub async fn #method_snake(&self, #short_args) -> #ret_ty {
             self.#method_with_opts(#short_delegate_args).await
         }
 
         #[doc = #doc_opts]
+        #extra_doc
         pub async fn #method_with_opts(&self, #opts_args) -> #ret_ty {
             #call_body
         }
--- a/connectrpc/src/client/mod.rs
+++ b/connectrpc/src/client/mod.rs
@@ -119,14 +119,40 @@ use buffa::view::MessageView;
 use buffa::view::OwnedView;
 use buffa::view::ViewReborrow;
 /// Re-export of [`futures::Stream`] (the `futures` 0.3 / `futures-core` 0.3
-/// trait), the request-stream bound for client-streaming calls. Re-exported
-/// so generated code and callers can name the bound without a direct
-/// `futures` dependency.
+/// trait), which [`ClientRequestStream`] builds on. Re-exported so generic
+/// code can name the trait without a direct `futures` dependency.
 pub use futures::Stream;
 /// Re-export of [`futures::stream::iter`]: adapts a collection that is
 /// already in hand into a request stream for a client-streaming call,
 /// without a direct `futures` dependency.
 pub use futures::stream::iter as stream_iter;
+
+mod sealed {
+    pub trait Sealed {}
+    impl<S> Sealed for S where S: super::Stream + Send + 'static {}
+}
+
+/// The request-stream bound for client-streaming calls.
+///
+/// Implemented automatically for every `Stream<Item = Req> + Send + 'static`
+/// — it cannot (and never needs to) be implemented by hand. The trait exists
+/// so the compiler can point at the two usual fixes when the bound is not
+/// met: wrap a ready collection with [`stream_iter`], and make a borrowing
+/// stream yield owned messages (the stream backs the request body, which can
+/// outlive the call frame and move across threads — hence `Send + 'static`).
+///
+/// Not to be confused with the server-side
+/// [`dispatcher::RequestStream`](crate::dispatcher::RequestStream), a boxed
+/// stream of raw request bytes.
+#[diagnostic::on_unimplemented(
+    message = "`{Self}` cannot be used as the request stream of a client-streaming call",
+    label = "expected an async `Stream<Item = {Req}> + Send + 'static`",
+    note = "for a collection that is already in hand, wrap it with `connectrpc::client::stream_iter(...)`",
+    note = "the stream backs the request body, so it must be `Send + 'static`: yield owned messages (no borrows of local data) or feed the call from a channel-backed stream"
+)]
+pub trait ClientRequestStream<Req>: sealed::Sealed + Stream<Item = Req> + Send + 'static {}
+
+impl<S, Req> ClientRequestStream<Req> for S where S: Stream<Item = Req> + Send + 'static {}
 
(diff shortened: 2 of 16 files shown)


iainmcgin:
[claude code] Benchmarked the stream-as-body design against the previous channel-based client-streaming (base `4da239d`) to answer whether the one-envelope-at-a-time body needs a buffer. Two independent criterion compare runs on a dedicated bare-metal host (c7i.metal-24xl, no shared tenancy), baseline and candidate binaries interleaved on the same box; untouched paths (`unary/small`, `server_stream`) as drift controls, which stayed ≤ ~1.7% throughout. Positive = this PR slower.

| bench | run 1 | run 2 |
|---|---|---|
| client_stream (10 small msgs) — connect / grpc / grpc-web | −2.0% / −2.2% / −2.0% | −1.2% / −1.6% / −0.9% |
| client_stream_many_small (1000 msgs) — connect / grpc / grpc-web | −3.5% / −2.0% / −3.2% | – |
| client_stream_large (10 × 1 MiB) — connect | **+9.4%** | **+6.1%** |
| client_stream_large — grpc | p=0.77 (±13%) | p=0.20 (±10%) |
| client_stream_large — grpc-web | −6.2% | +5.8% |

Reading:

- **Small-message streaming got 1–3.5% faster**, consistent across six benches and both runs: the per-message cross-task hop (bounded channel + wakeups + oneshot bridge) is gone, and the transport pulls frames directly. No buffering needed to keep the wire busy — HTTP/2 flow-control windows and socket buffers already do that.
- **1 MiB-message uploads over HTTP/1.1 + Connect regressed ~6–9%, reproducibly.** The magnitude matches lost encode/send overlap: each message costs ~2 MiB of encode+envelope memcpy that previously ran on the caller's task concurrently with the send task's socket writes, and now serializes on one task. The other two large benches are dominated by allocator bimodality at 1 MiB payloads (grpc-web flipped sign between runs; grpc never reached significance).
- The regression's substance is exactly the copy work already queued for 0.9.0: #219's refcount-chained framing removes the envelope memcpy for ≥16 KiB payloads, and the upstream encode work removes the encode-side copy. Re-running `client_stream_large` stacked on #219 should confirm most of it comes back.

The two new bench groups (`client_stream_many_small`, `client_stream_large`) are added in this PR so the comparison stays reproducible.

iainmcgin:
[claude code] Rebased onto main via merge: #226's abandonment fix targeted the channel-era detached send task, which this PR deletes — dropping the call future now drops the transport send and request body directly, so the guarantee holds structurally. All five #226 tests are ported to the stream API (the mid-drain backpressure case becomes an unread-request-body case, since there is no drain loop to park in), the `# Cancellation` rustdoc is kept with the mechanism description updated, `tokio/macros` stays consolidated on the base dependency (comment narrowed to the server accept loop), and the #226 changelog fragment is reworded to describe the behavior rather than the superseded mechanism. Full battery re-run post-merge: 562 all-features unit tests, Connect client conformance 2580/2580, lint/docs/MSRV/wasm/minimal clean.

connectrpc/src/client/mod.rs
```diff
@@ -3573,7 +3686,20 @@ where
...
         }
     })
-    .await
+    .await;
+
+    // An encode failure aborts the request at the transport level; surface
+    // the precise encode error instead of the generic transport failure —
+    // unconditionally, because a server's early response can race the abort
+    // and produce an `Ok` result for a truncated, encode-aborted upload.
+    if let Some(err) = encode_error
```
lovesegfault, line 3713: This covers the early-`Ok` race, but there's a second direction: with a background-driven body, an encode error can land *after* this check and is then never read — the call returns `Ok` for a truncated tail. I think that's fine (the server already finished its response), but a sentence saying it's intentional would help a future reader.

connectrpc-codegen/src/codegen.rs
```diff
@@ -2242,9 +2242,20 @@ fn generate_client_method(
...
     if client_streaming && !server_streaming {
         // Client-stream
+        extra_doc = quote! {
+            #[doc = ""]
+            #[doc = " `requests` is any `Stream<Item = ...> + Send + 'static` of"]
+            #[doc = " request messages (the `ClientRequestStream` bound); messages"]
+            #[doc = " are sent as the stream yields them. It backs the request"]
+            #[doc = " body, so yield owned messages or feed the call from a"]
+            #[doc = " channel-backed stream. For a collection that is already in"]
+            #[doc = " hand, wrap it with `::connectrpc::client::stream_iter(...)`."]
```
lovesegfault: Could we add a cancellation note here? This is the doc users actually read, and dropping the returned future (e.g. a `timeout` firing) now silently drops unsent messages. Something like: "Dropping the returned future cancels the call; unsent messages are never delivered." A matching sentence in the guide would be nice too.

connectrpc/src/client/mod.rs
```diff
@@ -118,6 +118,41 @@ use buffa::view::HasMessageView;
...
+}
+
+/// The request-stream bound for client-streaming calls.
+///
+/// Implemented automatically for every `Stream<Item = Req> + Send + 'static`
+/// — it cannot (and never needs to) be implemented by hand. The trait exists
+/// so the compiler can point at the two usual fixes when the bound is not
+/// met: wrap a ready collection with [`stream_iter`], and make a borrowing
+/// stream yield owned messages (the stream backs the request body, which can
+/// outlive the call frame and move across threads — hence `Send + 'static`).
```
lovesegfault, line 142: Worth a sentence here on panics: the stream is polled on the transport's task now, so a panic in `poll_next` surfaces as a generic transport error instead of propagating (and can take down a shared connection). Would save someone a confusing debug session.

connectrpc/src/client/mod.rs
```diff
@@ -2983,6 +3018,90 @@ impl Body for ChannelBody {
...
+    done: bool,
+}
+
+impl<S, Req> Body for EncodingBody<S>
+where
+    S: Stream<Item = Req>,
+    Req: buffa::Message + crate::codec::JsonSerialize,
+{
+    type Data = Bytes;
+    type Error = ConnectError;
```
lovesegfault, line 3067: Might as well add:

```rust
fn is_end_stream(&self) -> bool {
    self.done
}
```

The default is always `false`, so downstream layers need an extra `poll_frame` to notice exhaustion — `done` already means exactly this.

connectrpc/src/client/mod.rs
```diff
@@ -3469,22 +3618,32 @@ where
...
         config.compression_policy.with_override(options.compress),
     );
 
+    // The stream backs the request body directly: the transport polls it for
+    // the next frame as it is able to send. An encode failure is reported
+    // through the body (aborting the request) and stashed here so the call
+    // can surface the precise error instead of a generic transport failure.
+    let encode_error: std::sync::Arc<std::sync::Mutex<Option<ConnectError>>> =
+        std::sync::Arc::default();
+    let body: ClientBody = EncodingBody {
```
lovesegfault, line 3650: Nothing forces a construction site to also drain the mirror — the pairing is convention only. Maybe `EncodingBody::new(...) -> (Self, EncodeErrorSlot)` with a `#[must_use]` slot? Small change, makes the contract compiler-visible.

connectrpc/src/client/mod.rs
```diff
@@ -3429,27 +3548,57 @@ where
...
+/// Returns an error if a request message cannot be encoded, the transport
+/// fails, the whole-call deadline expires, the server responds with an
+/// error, or the response cannot be decoded.
 pub async fn call_client_stream<T, Req, RespView>(
     transport: &T,
     config: &ClientConfig,
     service: &str,
     method: &str,
-    requests: impl IntoIterator<Item = Req>,
+    requests: impl ClientRequestStream<Req>,
```
lovesegfault, line 3613: Style nit, take it or leave it: the other `call_*` fns name all their generics; `S: ClientRequestStream<Req>` would match them and show the bound in rustdoc. (Not about turbofish — that already works; naming it would just add a `_` to the conformance call.)

connectrpc/src/client/mod.rs
```diff
@@ -118,6 +118,41 @@ use buffa::view::HasMessageView;
...
 use buffa::view::OwnedView;
 use buffa::view::ViewReborrow;
+/// Re-export of [`futures::Stream`] (the `futures` 0.3 / `futures-core` 0.3
+/// trait), which [`ClientRequestStream`] builds on. Re-exported so generic
+/// code can name the trait without a direct `futures` dependency.
+pub use futures::Stream;
+/// Re-export of [`futures::stream::iter`]: adapts a collection that is
+/// already in hand into a request stream for a client-streaming call,
+/// without a direct `futures` dependency.
+pub use futures::stream::iter as stream_iter;
```
lovesegfault, line 128: Small one: we re-export `UnaryResponse`/`ServerStream`/`BidiStream` at the crate root, but not `stream_iter`, which most callers will want. `pub use client::stream_iter;` in lib.rs would shorten the common call site.

lovesegfault (commented):
Took a close look at the redesign — really nice work; no correctness or safety concerns from me. Comments below are all non-blocking: a few doc notes on correct-but-surprising behavior, plus small polish ideas.

connectrpc/src/client/mod.rs
iainmcgin: [claude code] Documented in f948aba. The comment now states that direction is deliberate and why: with the body driven in the background, an encode failure can land after the check and is then never read, so the call reports the server `Ok`. That is the intended outcome, because the server had already produced a complete response and the truncated tail did not affect it.

connectrpc-codegen/src/codegen.rs
iainmcgin: [claude code] Added in f948aba, on both the plain and `_with_options` generated methods, plus a matching paragraph in the guide (moved out of the code fence into the prose that follows it). The generated text calls out the `timeout` case specifically, since that is the shape most likely to hit it.

connectrpc/src/client/mod.rs
iainmcgin: [claude code] Added as a `# Panics in poll_next` section in f948aba. Worth recording that the blast radius is wider than it first looks: hyper eagerly polls the request body inline on the shared dispatch task before handing the pipe to a spawned task (`proto/h2/client.rs` in 1.9.0), so a panic on that first poll can fault a `SharedHttp2Connection` for every concurrent RPC on it, not just the one call. I deliberately did not put that mechanism in the doc, though, since it is a hyper internal that a refactor there would silently falsify. The doc states the durable part: the stream is polled on the task driving the request, a panic surfaces as a generic transport error, and where that task is shared it faults every RPC on the connection. It also notes the stream yields `Req` rather than a `Result`, so ending the stream early is the only way to report failure.

connectrpc/src/client/mod.rs
iainmcgin: [claude code] Checked this one against hyper rather than reasoning about it, and I think it would be a no-op, so I have left it out.

`done` is set in exactly two places, both inside `poll_frame`: when `poll_next` returns `None`, and in `record_error`. So it only becomes true as a result of the very poll that already returned `None` or an error, never before it.

Hyper checks `is_end_stream()` immediately after receiving a data frame, to decide whether to set END_STREAM on it (`proto/h2/mod.rs:158` in 1.9.0, `let is_eos = me.stream.is_end_stream();` right after `poll_frame` yields data). At that moment `done` is still false, so the optimization would not fire and hyper would poll once more anyway, which is the extra poll the suggestion aims to avoid. The other call site (`proto/h2/client.rs:712`) runs before the first poll, where `done` is false by construction.

Getting the actual saving needs one-item lookahead so the body knows the stream is exhausted while returning the last frame, which is a design change rather than a three-line addition. Happy to file that as a follow-up if you think the saved frame is worth it.

connectrpc/src/client/mod.rs
iainmcgin: [claude code] Agreed the pairing is convention-only. I have deferred it rather than doing it here: there is exactly one construction site today, and unlike the rest of this round it is a refactor rather than a doc change, so it would mean re-running the verification battery on an already-verified PR. Filing it as a follow-up issue so the constraint gets made compiler-visible before a second construction site appears, which is when it would actually start costing something.

connectrpc/src/client/mod.rs
iainmcgin: [claude code] Left as is. Naming the generic would show the bound in a where-clause, but it also adds a type parameter to the turbofish arity, so the existing call sites need an extra `_` as you note. That felt like a poor trade for a cosmetic gain, even in a release where we are already breaking things. Happy to take it if you feel strongly.

connectrpc/src/client/mod.rs
iainmcgin: [claude code] Done in f948aba, and it turned out to be the most valuable comment of the batch. Adding the root re-export exposed that we were teaching two spellings: the guide and the unmet-bound diagnostic said one thing, while the generated method docs, the `call_client_stream` example, the streaming-tour example and its README, and the streaming e2e test all still said `connectrpc::client::stream_iter`. So a caller who hit the compile error, followed the compiler hint, then hovered the generated method would read a different path than the one they had just typed. Everything is now unified on `connectrpc::stream_iter`; the only remaining mention of the long path is the re-export declaration itself. Verified end-to-end with a live client-streaming call over a socket through the root path.

lovesegfault (approved):

PR #228, opened 2026-07-12
client: add BidiStream::into_split for independently owned halves

iainmcgin:
## Summary

Adds `BidiStream::into_split`, splitting a bidirectional stream into independently owned halves so the two sides can be driven from separate tasks — true full duplex:

```rust
let (mut send, mut recv) = client.running_sum().await?.into_split();
let reader = tokio::spawn(async move {
    while let Some(reply) = recv.message().await? { /* ... */ }
    Ok::<_, connectrpc::ConnectError>(())
});
for req in requests {
    send.send(req).await?;
}
send.close_send();
reader.await.expect("reader task")?;
```

Builds on #227, now merged, and rebased onto `main`; additive — no existing API changes, no codegen changes (generated bidi methods still return `BidiStream`).

## Design

`BidiStream` was already two disjoint sides internally; the split is a plain destructure with no locking added:

- **`BidiSendHalf<Req>`** owns `send`/`close_send` (and only needs `Req` bounds — looser than the combined impl). It carries its own copy of the whole-call deadline. Dropping it ends the request body cleanly; the RPC continues until the receive half finishes.
- **`BidiRecvHalf<B, RespView>`** owns `message`/`headers`/`trailers`/`error` and the lazy response-initialization state machine. The abort-on-drop of in-flight initialization tasks (#221) moves here from `BidiStream`'s `Drop` — which also frees `BidiStream` of its `Drop` impl so `into_split` can destructure. Dropping this half cancels the RPC, exactly as dropping a whole `BidiStream` did; a code comment records that the `send`-before-`recv` field order is load-bearing for the drop sequence.
- `BidiStream` keeps its full API via thin delegation, and `into_split` sits in an unbounded impl (a pure move needs no bounds). No `reunite` — documented on both halves.
- Docs state the HTTP/2 requirement for interleaved use prominently on `into_split` itself (an HTTP/1.1 caller doing response-dependent sends would deadlock), and steer users toward spawned tasks over naming the halves' body type parameter.

## Testing

- Three new e2e tests over a real gRPC/h2 connection: full-duplex ping-pong with the halves owned by different tasks (each echo received before the next send), dropping the send half ends the stream cleanly, dropping the receive half fails subsequent sends instead of hanging.
- All 562 all-features unit tests pass unchanged through the delegating methods, including the #221 cancellation-safety tests.
- Live socket drive: split ping-pong over h2 completes in ~50 ms alongside the existing client-stream probes.
- Conformance client suites re-run: Connect 2580/2580; gRPC shows only the known environment-flaky Timeouts cases (including unary cases this diff does not touch).
- `cargo semver-checks` vs 0.8.1: 196/196 (additive). Net non-test size: +177 lines.

### DIFF
--- a/.changes/unreleased/Added-20260712-110850.yaml
+++ b/.changes/unreleased/Added-20260712-110850.yaml
@@ -1,0 +1,10 @@
+kind: Added
+body: |-
+  **`BidiStream::into_split`** splits a bidirectional stream into
+  independently owned `BidiSendHalf` and `BidiRecvHalf`, so the two sides
+  can be driven from separate tasks (true full duplex). The split is a plain
+  move of the stream's two sides — no locking is added. Dropping the send
+  half (or calling `close_send`) ends the request body cleanly while the
+  RPC continues; dropping the receive half cancels the RPC, matching the
+  behavior of dropping a whole `BidiStream`.
+time: 2026-07-12T18:08:50.214607301Z
--- a/connectrpc/src/client/mod.rs
+++ b/connectrpc/src/client/mod.rs
@@ -3003,8 +3003,8 @@ where
 /// A request body that pulls envelope-encoded frames from an mpsc channel.
 ///
 /// Used as the request body for bidirectional streaming calls.
-/// [`BidiStream::send`] pushes encoded envelopes to the channel's sender half;
-/// dropping the sender (via [`BidiStream::close_send`]) closes the body,
+/// [`BidiSendHalf::send`] pushes encoded envelopes to the channel's sender;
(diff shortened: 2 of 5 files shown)


iainmcgin:
[claude code] Restacked onto the current #227 tip (`f948aba`), which picked up that PR review round. Rebase was clean — no conflicts with the doc and re-export changes there, despite both touching `client/mod.rs`. Branch is now `8950ea0`; `task lint` and the full workspace suite (56 suites, 0 failures) are green on the restacked branch.

This is now targeted at the 0.9.0 release alongside #227, so it needs to land right after its parent.

azdagron (approved):

PR #240, opened 2026-07-23
error: preserve the source chain of client transport failures

### LINKED ISSUE #237: ConnectError discards the transport error's source chain across every client transport path
Related to #93, which proposes adding a `source` field to `ConnectError` from the handler-side error-conversion angle. This issue is the client-transport-side evidence for the same gap.

## The gap

`ConnectError`'s `Error` impl never overrides `source()`, so it always returns `None`:

https://github.com/connectrpc/connect-rust/blob/8b3c3b05d3b54af547477a9e3b3a77d62f68e229/connectrpc/src/error.rs#L482

The HTTPS client transport path converts a transport failure into `ConnectError` via `format!("HTTPS request failed: {e}")` — `Display` only:

https://github.com/connectrpc/connect-rust/blob/8b3c3b05d3b54af547477a9e3b3a77d62f68e229/connectrpc/src/client/mod.rs#L817-L819

`e`'s own `source()` chain (the actual DNS failure / connection-refused / connection-reset / timeout cause) is discarded right here, before any caller code runs. The plaintext path does the same thing (`client/mod.rs#L800`), and the same pattern recurs in the HTTP/2 transport (`client/http2.rs`, unix-socket connect / eager connect / h2 send / TLS handshake) and in `From<std::io::Error> for ConnectError` (`error.rs#L484-L488`) — one conversion pattern, repeated at every place a transport error becomes a `ConnectError`.

`map_transport_send_error` (`client/mod.rs#L226-L232`, since #199) only preserves a `ConnectError` already present in an error's chain; its fallback for everything else is still `Display`-only, so it doesn't help at any of these sites.

## Impact

Any caller of this client gets the exact same one-line message for any connect-stage failure, regardless of transport — a DNS failure, a refused connection, a reset, and a timeout are all indistinguishable in the returned `ConnectError`. This can't be fixed by logging more carefully downstream: the detail is gone before the caller ever sees the error.

## Suggested fix

Same shape as #93's proposal: a non-serialized local field on `ConnectError` (e.g. `source: Option<Box<dyn std::error::Error + Send + Sync>>`), populated at these conversion sites alongside the existing `Display`-formatted `message` (unchanged, so wire format and `Display` output stay identical), returned from `Error::source()`. Purely additive, not a breaking change.

christopherwxyz:
## Summary

`ConnectError`'s `Error` impl never overrode `source()`, so any client-side transport failure — a DNS failure, a refused/reset connection, a TLS handshake failure, a timeout — was reduced to a single `Display`-formatted string before any caller code ran. Fixes #237.

- Adds a non-serialized `source: Option<Arc<dyn std::error::Error + Send + Sync>>` field to `ConnectError`, a `ConnectError::with_source(..)` builder, and an `Error::source()` override that returns it.
- `Arc` rather than the `Box` suggested in #237, specifically so `ConnectError` keeps its existing `#[derive(Clone)]` without requiring the wrapped error to be `Clone`.
- `message`/`Display` output — and therefore wire format — is unchanged; this is purely additive.
- Populates `source` at every place a transport error becomes a `ConnectError`, named in the issue:
  - `HttpClient::send`'s plaintext and TLS paths
  - `map_transport_send_error`'s fallback branch
  - `Http2Connection`: unix-socket connect, TLS handshake, eager/lazy connect (`drive_connect`), h2 send, and (bonus) the establishment-timeout branch
  - `From<std::io::Error>` and `From<http::Error>` for `ConnectError`
- A shared `unavailable_from_transport_error` helper collapses the repeated `format!` + `.with_source(..)` idiom across `client/mod.rs` and `client/http2.rs`, boxing the error once and reusing it for both the message and the source.

## Docs

`docs/guide.md` gets a short paragraph on `.with_source(..)`/`Error::source()`, including that it's local-only — a `ConnectError` decoded from a server's response never has one.

## Test plan

- [x] Added/extended unit tests covering every named conversion site (`HttpClient` HTTP/HTTPS, unix-socket connect, TLS handshake, eager connect, h2 send, establishment timeout, `map_transport_send_error` fallback, both `From` impls) — all assert `Error::source()` is populated.
- [x] `cargo build -p connectrpc --all-features`
- [x] `cargo clippy -p connectrpc --all-features --all-targets` with `-D warnings`
- [x] `cargo test -p connectrpc --all-features` — 584/584 lib tests pass, 6/6 doctests pass (13 pre-existing tests in the unrelated `max_connection_age`/`max_connection_idle` graceful-shutdown family were excluded — confirmed flaky/hanging on a clean `main` checkout as well, unrelated to this change)
- [x] Changelog fragment added under `.changes/unreleased/`

Co-authored with Claude via research on idiomatic Rust 1.95+ error-source patterns.

### DIFF
--- a/.changes/unreleased/Added-20260723-023643.yaml
+++ b/.changes/unreleased/Added-20260723-023643.yaml
@@ -1,0 +1,16 @@
+kind: Added
+body: |-
+  **`ConnectError` now preserves the underlying cause of transport
+  failures.** A new `ConnectError::with_source` builder attaches an
+  underlying error, surfaced through `Error::source()`; `ConnectError` stays
+  `Clone` since the source is held in an `Arc`. Every place a transport
+  error was previously stringified into `message` and discarded now
+  retains it: the HTTP and HTTPS `HttpClient` transports, `Http2Connection`
+  (unix-socket connect, eager/lazy connect, TLS handshake, h2 send), and
+  `From<std::io::Error>` / `From<http::Error>` for `ConnectError`. The wire
+  format and `Display` output are unchanged; the derived `Debug` output now
+  additionally includes the source when one is attached. This is purely
+  additive ([#237]).
+
+  [#237]: https://github.com/connectrpc/connect-rust/issues/237
+time: 2026-07-23T02:36:43.000000000Z
--- a/.changes/unreleased/Added-20260723-023643.yaml
+++ b/.changes/unreleased/Added-20260723-023643.yaml
@@ -2,15 +2,21 @@ kind: Added
 body: |-
   **`ConnectError` now preserves the underlying cause of transport
   failures.** A new `ConnectError::with_source` builder attaches an
-  underlying error, surfaced through `Error::source()`; `ConnectError` stays
-  `Clone` since the source is held in an `Arc`. Every place a transport
-  error was previously stringified into `message` and discarded now
-  retains it: the HTTP and HTTPS `HttpClient` transports, `Http2Connection`
-  (unix-socket connect, eager/lazy connect, TLS handshake, h2 send), and
-  `From<std::io::Error>` / `From<http::Error>` for `ConnectError`. The wire
-  format and `Display` output are unchanged; the derived `Debug` output now
-  additionally includes the source when one is attached. This is purely
-  additive ([#237]).
+  underlying error, surfaced through `Error::source()` and, as a
+  `SharedSource` handle that can be moved into another error type, through
+  `ConnectError::source_arc()` (with `with_shared_source` to re-attach one);
+  `ConnectError` stays `Clone` since the source is held in an `Arc`. The
+  built-in client transports now retain the cause wherever a failure they
+  classify was previously stringified into `message` and discarded: the HTTP
+  and HTTPS `HttpClient` transports, `Http2Connection` (unix-socket connect,
+  eager/lazy connect and its establishment timeout, TLS handshake, h2 send),
+  and `From<std::io::Error>` / `From<http::Error>` for `ConnectError`.
+  `ConnectError::unavailable_from_transport` exposes the same convention to
+  custom `ClientTransport` implementations. Errors the call path synthesises
+  itself (the call deadline, request build/encode, response decode, a
+  body-read reset) do not carry one. The wire format and `Display` output are
+  unchanged; the derived `Debug` output now additionally includes the source
+  when one is attached. This is purely additive ([#237]).
 
   [#237]: https://github.com/connectrpc/connect-rust/issues/237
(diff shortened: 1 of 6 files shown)


iainmcgin:
[claude code] Thanks for this — the core change is right and carefully done (the `Arc`-over-`Box` choice, `#[serde(skip)]`, and preserving every message byte-for-byte all check out). To get it into 0.9.0 I've rebased the branch onto current `main` (dropping the merge commit; your two commits are otherwise unchanged) and pushed one maintainer commit on top rather than round-tripping review comments.

What the fixup changes:

**Two accessors the review found missing.** `Error::source()` lets a caller inspect the cause but not carry it, and the whole point of holding it in an `Arc` was shareability — so there's now `ConnectError::source_arc() -> Option<SharedSource>` (a `pub type SharedSource = Arc<dyn Error + Send + Sync>` alias) and `with_shared_source(SharedSource)` for re-attaching one when rebuilding an error under a different code. The `SharedSource` doc records a std wart worth knowing: `Arc<T>` implements `Error` (unlike `Box<T>`), so a `#[source] SharedSource` field in a thiserror enum shows up in that enum's chain as an opaque `Arc` link — downcast the handle directly. The `source_arc` doctest demonstrates that.

**Your `unavailable_from_transport_error` helper is now public** as `ConnectError::unavailable_from_transport`, so a custom `ClientTransport` can follow the same convention the built-in ones now do; `ClientTransport::Error`'s doc points at it, and its own doc warns against wrapping an already-classified `ConnectError` (which would bury the code under `unavailable`).

**Docs narrowed to what the code does.** The guide paragraph and changelog said transport failures "(DNS, connection refused/reset, TLS handshake, timeout)" populate the source. Two of those don't: a reset while reading the response body goes through `classify_body_read_error`, whose error is bound only by `Display` on the public `call_*` signatures (tightening that is a breaking change for another day), and the per-call deadline is synthesised by the call path with nothing to attach. Both now say a source is present when the *transport* classified the failure and absent when the call path did, and mention that `message` deliberately repeats the cause's `Display` text (so `anyhow`'s `{:#}` prints it twice).

**Tests:** `source` is neither serialized nor deserialized (the one wire-visible risk rested on a single attribute with no test); `source_arc` downcasts after the owning error is dropped; `with_shared_source` keeps the chain link transparent; `unavailable_from_transport` keeps the cause in both message and source.

Also reworded the `find_connect_error_in_chain` comment (its "boxed trait objects cannot appear as links" premise needed qualifying now that an `Arc`'d source is a link) and restored the establishment-timeout message text to `main`'s while keeping `Elapsed` as the source.

646 unit tests + the new doctest pass; clippy and rustdoc `-Dwarnings` clean.

rpb-ant (approved):

PR #257, opened 2026-07-24
client: resolve the gRPC unary response status before the HTTP status

### LINKED ISSUE #241: client: unary gRPC calls lose a proxy's trailers-only error message and metadata
## Summary

A trailers-only gRPC error response is decoded differently by the unary client and the streaming client, because the two paths check HTTP status and `grpc-status` in the opposite order. The unary path loses the server's error message and its metadata.

## The divergence

`parse_grpc_unary_response` checks the HTTP status first and returns immediately, so `grpc-status` is never consulted:

```rust
// connectrpc/src/client/mod.rs, parse_grpc_unary_response
let status = response.status();
let resp_headers = response.headers().clone();

// Non-200 HTTP status (connection-level error)
if !status.is_success() {
    let code = http_status_to_error_code(status);
    let mut err = ConnectError::new(code, format!("HTTP error {}", status.as_u16()));
    err.set_response_headers(resp_headers);
    return Err(err);
}
```

`make_server_stream` checks the trailers-only response first:

```rust
// connectrpc/src/client/mod.rs, make_server_stream
// For gRPC, check for trailers-only error response
if matches!(protocol, Protocol::Grpc | Protocol::GrpcWeb)
    && let Some(mut err) = parse_grpc_error_from_trailers(&response_headers)
{
    err.set_response_headers(response_headers);
    return Err(err);
}

// Non-200 responses are protocol errors
if !status.is_success() {
```

## Why it matters

The case this affects is a proxy synthesizing a trailers-only error: an Envoy local reply carrying `HTTP 503` together with `grpc-status: 14` and `grpc-message: upstream connect error`. The same bytes produce:

- server-streaming: `Unavailable("upstream connect error")`, with the server's trailer metadata attached
- unary: `Unavailable("HTTP error 503")`, with the message and metadata discarded

Unary is the most common RPC shape, and running behind a proxy is the most common deployment, so the worse diagnostic lands in the more common case. The codebase already treats this as a case worth handling — `validate_grpc_response_content_type`'s doc comment names "proxies that synthesize trailers-only error responses (such as Envoy local replies)".

Note that the HTTP status code itself is not lost, because `http_status_to_error_code(503)` and the `grpc-status: 14` both map to `Unavailable`. What is lost is the message and the metadata, which is the part a human needs when debugging.

## Proposed fix

Hoist the trailers-only check above the HTTP-status check in `parse_grpc_unary_response`, so both paths run one sequence: trailers-only, then status, then content-type, then encoding support.

## Notes

No conformance suite covers this. All six suites drive a well-behaved reference server, which never emits a non-200 response carrying `grpc-status`; the case only arises with an intermediary. Needs unit coverage.

### LINKED ISSUE #242: client: gRPC-Web trailer-frame check is ungated, so plain gRPC bodies can clobber real trailers
## Summary

`parse_grpc_unary_response` checks for a gRPC-Web trailer frame in two places. One is gated on `Protocol::GrpcWeb`; the other, 36 lines later in the same function, is not. On plain gRPC a response body whose first byte has the high bit set is therefore parsed as a gRPC-Web trailer block, and silently overwrites any HTTP/2 trailers already captured.

## The divergence

Gated, in the body-read loop:

```rust
// connectrpc/src/client/mod.rs, parse_grpc_unary_response
if matches!(config.protocol, Protocol::GrpcWeb)
    && let Some(trailer_end) = grpc_web_trailer_frame_end(&buf)
{
    buf.truncate(trailer_end);
    break;
}
```

Ungated, in the envelope-scan loop of the same function:

```rust
while !buf.is_empty() {
    // Check for gRPC-Web trailer frame (flag 0x80)
    if buf[0] & 0x80 != 0 {
        let decompression = response_encoding
            .as_deref()
            .map(|enc| (&config.compression, enc));
        if let Some(trailers) =
            parse_grpc_web_trailer_frame_with_compression(&buf, decompression)
        {
            grpc_trailers = trailers;
        }
        break;
    }
```

The streaming path gates both of its equivalent sites, and states why the gate is load-bearing:

```rust
// For gRPC-Web, check for a complete trailer frame (flag 0x80)
// before attempting envelope decode (which would treat 0x80 as
// a data envelope flag rather than the gRPC-Web trailer sentinel).
```

## Why it matters

`0x80` is the gRPC-Web trailer sentinel, and it is also a legal-looking value of the same flag byte in a normal envelope header. `Envelope::decode_with_limit` does not validate unknown flag bits, and `envelope::flags` defines only `DATA` (`0x00`), `COMPRESSED` (`0x01`) and `END_STREAM` (`0x02`) — `0x80` is absent from the module entirely, which is why every site that needs it hand-rolls the literal.

So on `Protocol::Grpc`, a first byte with the high bit set takes the trailer-frame branch, and `grpc_trailers = trailers` replaces trailers that were legitimately captured from the HTTP/2 trailer section. The same bytes over server-streaming-on-gRPC are handled correctly.

## Proposed fix

Add the `Protocol::GrpcWeb` gate to the second site. On `Protocol::Grpc`, treat a set high bit as a framing error (`internal`) rather than as data, since it is not a defined flag in that protocol.

While here it is worth adding `GRPC_WEB_TRAILER: u8 = 0x80` to `envelope::flags`, so the sentinel stops being a bare literal — it currently appears hand-written at five sites across `service.rs` and `client/mod.rs`.

## Notes

Not covered by conformance: a conformant gRPC server never sets the high bit, so the misparse needs a hostile or broken peer to trigger. Needs unit coverage.

iainmcgin:
Fixes #241
Fixes #242

Two divergences between the unary gRPC response parser and its streaming sibling.

**#241 — a proxy's trailers-only error lost its message and metadata on unary calls.** The unary path checked the HTTP status first and returned immediately, so `grpc-status` was never consulted. Given an Envoy local reply carrying `503` alongside `grpc-status: 14` and `grpc-message: upstream connect error`, a server-streaming call reported `Unavailable("upstream connect error")` with the server's metadata and a unary call reported `Unavailable("HTTP error 503")` with both discarded.

The unary path now resolves the gRPC status whenever the reply plausibly speaks gRPC, and keeps the HTTP-derived code otherwise. The header status is honoured only for a genuine trailers-only response, tested as `!has_body_data`.

**#242 — on plain gRPC, a high bit in the first body byte could overwrite the real trailers.** The gRPC-Web trailer-frame check appears twice in this function and only the first was gated on the protocol. `0x80` is the gRPC-Web trailer sentinel and is not a defined flag in plain gRPC, but `Envelope::decode_with_limit` does not validate unknown flag bits, so the frame was parsed and `grpc_trailers` was replaced — including over trailers already captured from the HTTP/2 trailer section. Both sites are gated now, and the sentinel gets a name.

## Verified against real Go implementations

Both changes are wire-visible interpretation, and neither case is reachable through conformance — the suite has no non-200 raw-response case and no conformant server emits `0x80`. So I built a raw-h2c harness emitting each shape exactly and drove **connect-go v1.19.1**, **grpc-go v1.81.1**, this branch and its base against identical bytes.

**The premise in #241 was half wrong, in our favour.** connect-go does check the HTTP status first — `grpcValidateResponse` returns `httpToCode(status)` before touching content-type or any gRPC status. **grpc-go does not**: a valid gRPC content-type sets `isGRPC` and the HTTP-status branch is never entered. Measured on the Envoy shape, grpc-go returns `Unavailable("upstream connect error")` and so does this branch, byte for byte; on `500` + `grpc-status: 3` both return `InvalidArgument("bad argument")`. It is the *base* tree that matched connect-go. So this matches grpc-go and diverges from connect-go, on a case where connect-go discards the server's code, message and metadata entirely.

**On `0x80`, both Go clients already reject the frame** — connect-go with "invalid envelope flags 128", grpc-go with "received unexpected payload format 128". The base tree was the outlier: it parsed the attacker's frame as gRPC-Web trailers and reported `NotFound`, *overriding real HTTP/2 trailers that said `grpc-status: 0`*. This removes a divergence rather than creating one.

**No incompatibility is introduced.** Across fourteen scenarios there is no case where a Go client completes a call and this branch fails it. Every difference is a different error for a call that fails in all four clients.

The interop run also caught a regression in an earlier revision of this PR: a 502 carrying an HTML page reported `Unknown` where base, connect-go and grpc-go all report `Unavailable`, losing the retryability signal. Fixed, and the rule is now applied wherever a body fails to parse, so the two spellings of the same proxy failure cannot produce different retry decisions.

One pre-existing divergence found and deliberately not adopted: grpc-go accepts a *successful* response delivered on HTTP 503. We and connect-go reject it, which seems right.

## Conformance

The trailers-only gate is load-bearing and an earlier revision got it wrong — hoisting the header check above the body read fails four cases named `trailers-only/ignore-header-if-body-present`, measured at 12 passed / 4 failed versus 16/16 for this one. Full suites: gRPC-Web client 2838/0, gRPC client 1452/1454 with both failures in the load-induced `Timeouts` flake that is equally present on base, and the `gRPC Unexpected Responses` subset 22/0.

One note on CI: two `handler::tests` element-budget tests fail on `main` right now, independently of this change — fixture rot from buffa 0.9.1, fixed by #239.

### DIFF
--- a/.changes/unreleased/Fixed-20260723-222329.yaml
+++ b/.changes/unreleased/Fixed-20260723-222329.yaml
@@ -1,0 +1,42 @@
+kind: Fixed
+body: |-
+  **A gRPC or gRPC-Web unary call now reports a trailers-only error instead of
+  the bare HTTP status** ([#241]). When a proxy answers with a non-200 status
+  and a gRPC status in the response headers — an Envoy local reply sends `503`
+  alongside `grpc-status: 14` and `grpc-message: upstream connect error` — the
+  call surfaces the server's code, message, and metadata rather than
+  `HTTP error 503`. This matches grpc-go, which decides that a reply is gRPC
+  from its content-type and reads the status from it; connect-go instead
+  reports the HTTP status and discards the server's code, message and metadata
+  entirely. The header status applies only to a genuine trailers-only
+  response: a response carrying body data or HTTP/2 trailers takes its status
+  from those, never from the headers, and a success delivered on a non-200 is
+  still refused. The error code can change as a result, not just the message:
+  a proxy pairing HTTP 500 with `grpc-status: 3` now yields `InvalidArgument`
+  where it previously yielded `Unknown`, so a caller matching on `ErrorCode`
+  may observe the server's code rather than the one derived from the HTTP
+  status.
+
+  A non-2xx reply that is not a usable gRPC response reports its HTTP status,
+  so an nginx or load-balancer error page still arrives as `Unavailable` on a
+  `502` and stays recognizable as retryable. What went wrong is appended to
+  the message rather than replacing the code, and this now holds however the
+  error page fails: `HTTP error 502: unexpected content-type: text/html` when
+  the proxy labels it as HTML, and `HTTP error 502: envelope decode failed:
+  ...` when the proxy keeps the gRPC content-type and the page fails framing
+  instead. A reply with no `content-type` at all is still parsed for a
+  `grpc-status`, since an absent content-type is not evidence the peer is not
+  speaking gRPC.
+
+  **A plain gRPC unary response whose body sets the `0x80` envelope flag is
+  now a framing error** ([#242]). That flag marks a gRPC-Web trailer frame
+  and is not defined for gRPC, so a server or intermediary that set it could
+  replace the trailers received in the HTTP/2 trailer section with trailers
+  of its own choosing, including the `grpc-status`. gRPC-Web is unaffected:
+  its trailer frames are parsed as before. The flag now has a name,
+  `envelope::flags::GRPC_WEB_TRAILER`, public alongside the existing `DATA`,
+  `COMPRESSED` and `END_STREAM` flags.
+
+  [#241]: https://github.com/connectrpc/connect-rust/issues/241
+  [#242]: https://github.com/connectrpc/connect-rust/issues/242
+time: 2026-07-23T22:23:29.547732463-07:00
--- a/connectrpc/src/client/mod.rs
+++ b/connectrpc/src/client/mod.rs
@@ -255,7 +255,7 @@ fn grpc_web_trailer_frame_end(data: &[u8]) -> Option<usize> {
         if frame_end > data.len() {
             return None;
         }
-        if data[offset] & 0x80 != 0 {
+        if data[offset] & crate::envelope::flags::GRPC_WEB_TRAILER != 0 {
             return Some(frame_end);
         }
         offset = frame_end;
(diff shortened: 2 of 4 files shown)


iainmcgin:
The conformance gap this PR relies on is now filed upstream as connectrpc/conformance#1119 (non-200 carrying a gRPC status — zero cases pair the two, and connect-go and grpc-go disagree) and connectrpc/conformance#1120 (reserved envelope flag bytes).

#1119 is the one that bears on reviewing this PR: it records that the divergence is between the two reference implementations rather than something we invented, with the measured behaviour of each and the mechanism in their source. If upstream settles the question one way, this PR's choice should follow it.

asacamano (approved):

PR #262, opened 2026-07-24
client: carry response metadata on every terminal error

### LINKED ISSUE #202: client: attach response headers/trailers consistently to terminal RPC errors
## Summary

Terminal RPC errors surfaced by the client attach response headers and trailers inconsistently across stream shapes and protocols. #192 and #201 made the malformed-END_STREAM parse error carry response headers on both the client-streaming and server-streaming Connect paths, but the review of #201 inventoried several remaining sites where equivalent errors drop metadata that the sibling path preserves.

## Inventory (line numbers as of main after #201)

All in `connectrpc/src/client/mod.rs`:

1. **Carried END_STREAM error, server-streaming path** — `process_end_stream` converts a server error in a *well-formed* END_STREAM via `end_stream_error_to_connect_error(err)` with neither headers nor trailers attached (the `end_stream.error` match arm). The client-streaming equivalent attaches **both** (`set_response_headers` + `set_trailers` before returning). This is the most user-visible gap: it is the normal error path for a failed Connect server-streaming RPC, and once the error is `?`-propagated away from the stream object the metadata is unreachable (`ServerStream::headers()`/`trailers()` only help while the stream is still held).
2. **END_STREAM decompression failure** — drops headers on *both* shapes (`process_end_stream`'s `decompress_envelope` arm, and the client-stream `map_response_decompression_error` site). Symmetric today; whatever rule is adopted should cover both.
3. **gRPC / gRPC-Web streaming terminal errors** — `classify_grpc_end` constructs its error variants without attaching `self.headers`, and `parse_grpc_error_from_trailers` attaches nothing, while the unary gRPC path attaches `resp_headers` to every error site.
4. **Connect unary content-type rejection** — the Connect-protocol unary parse path's content-type error does not call `set_response_headers`, while the gRPC unary equivalent (after #200) does.
5. **Mid-stream body-read errors** — wrapped as `internal` (`"failed to read response body: ..."`) at three sites, flattening any classified error a wrapped response body might carry; fixing this needs a `'static` bound on the body error type, so it may warrant its own decision.

## Proposed rule

A terminal `ConnectError` returned to the caller carries the response headers whenever they were received, and the trailers whenever they were parsed, regardless of stream shape or protocol. Document the rule on `ConnectError::response_headers` once adopted.

## Notes

Surfaced by the dual review passes on #201 (and #199/#200 for items 4–5). Sites 1–4 are each the same few-line pattern as the #201 fix; item 5 is the only one with an API-bound implication.

iainmcgin:
Fixes #202

Terminal client errors now carry the response headers whenever they were received, and the trailers whenever they were parsed, regardless of stream shape or protocol.

The inventory in #202 listed four sites where an error dropped metadata its sibling path preserved. Rather than adding `set_response_headers` at each, the metadata is attached where the terminal record is written, so the rule holds by construction: `ServerStream::message()` has the only write, and every `StreamEnd` funnels through it — carried END_STREAM errors, decompression failures, the `classify_grpc_end` variants and the `parse_grpc_error_from_trailers` results. The explicit call added by #201 is now redundant and removed. Writing the metadata in rather than applying it on read matters because `ServerStream::error()` reads the stored record directly. The remaining bare sites on the Connect unary and client-streaming paths are covered too, so "consistently" means what it says.

Item 5 in the issue is untouched; it needs a `'static` bound on the body error type and is tracked as #237.

**The choke point initially clobbered curated trailers.** `parse_grpc_error_from_trailers` deliberately excludes the gRPC status keys from error metadata, and attaching the raw map on top re-exposed them. Guarding on "already set" would have been unsound — trailers carrying only `grpc-status` produce a legitimately empty curated map — so the filter is extracted and the choke point recomputes it, which is idempotent on the gRPC path and a no-op on Connect.

## The server-side half, and why it is here

Attaching upstream headers to streaming errors would have widened a framing bug that already existed on the unary paths, so it is fixed rather than shipped and filed.

Our server echoed `err.response_headers()` onto the wire with a builder that *appends*. Driven against **connect-go v1.19.1** and **grpc-go v1.81.1**, with a per-header bisect, that is not hygiene — it is a hard break. A gateway propagating an upstream failure over Connect on HTTP/1.1 emits a response hyper cannot serialize at all: **the connection closes with zero bytes written**, connect-go reports `unexpected EOF` with empty metadata, and a handler's `permission_denied` reaches the caller as `unavailable`. Every error code collapses the same way, because all the client ever sees is a transport failure. `content-length` is the sole culprit. An echoed `content-encoding` is a second, milder break — connect-go tries to gunzip the plain-JSON error body and falls back to the HTTP status — and an echoed `date` displaces the one hyper would otherwise set, so the response carries a false generation time.

`echo_error_headers` applies two rules: a header the response already set wins, and names that would describe something other than this response are dropped — hop-by-hop, body framing, provenance, status-bearing. Legitimate metadata is unaffected: `x-upstream-region` and **both** values of a multi-valued `x-upstream-tag` arrive at every client on every shape, which is the case the snapshot-before-append logic could plausibly have broken.

Dropping `grpc-encoding` was checked rather than assumed. grpc-go only consults it when a frame's compressed flag is set, and these responses have no data frames, so an echoed one is inert today — confirmed, including under `AcceptCompressors("identity")`, which makes grpc-go police the advertised encoding. It becomes live the moment a response carries both the echoed encoding and a message frame, and the Connect analogue of the same class already is.

**Trailing metadata is still echoed unfiltered.** The same class exists one frame later, but fixing it needs a policy rather than a predicate — provenance is not derivable where the filtering happens, the obvious signal is remotely drivable, and the source-selection branch has the same bug one level up. An attempt to fix it here was reverted for those reasons and is now #263.

## Verification

Conformance: server 3600/0, server Connect+TLS 2396/0, client Connect 2580/0, client gRPC-Web 2838/0. The gRPC client suite fails 1-4 cases in `Timeouts` across four runs with a different set each time; base flakes identically, including one clean run.

`duplicate_metadata.yaml` is the case worth naming — it echoes multi-valued `x-custom-header` onto error responses across all four RPC kinds. Conformance never asks a server to echo a connection-scoped header, though, and never echoes a name the server also sets, so neither new rule is exercised by the suite; unit tests and the interop harness cover those.

One note on CI: two `handler::tests` element-budget tests fail on `main` right now, independently of this change — fixture rot from buffa 0.9.1, fixed by #239.

### DIFF
--- a/.changes/unreleased/Fixed-20260723-222414.yaml
+++ b/.changes/unreleased/Fixed-20260723-222414.yaml
@@ -1,0 +1,24 @@
+kind: Fixed
+body: |-
+  **Terminal client errors now carry the response headers and trailers
+  consistently** ([#202]). A failed RPC reported through
+  `ServerStream::message()` — a Connect END_STREAM carrying an error, a gRPC
+  or gRPC-Web `grpc-status` error, a decode, decompression, transport or
+  deadline failure — arrives with `ConnectError::response_headers()`
+  populated, and with `ConnectError::trailers()` populated whenever
+  termination metadata was received. The same holds for the sticky replay on
+  a second `message()` call and for `ServerStream::error()`. Previously only
+  a few of these paths attached metadata, so whether a caller could read the
+  server's headers off an error depended on which protocol was in use and on
+  how the stream happened to fail. The Connect unary and client-streaming
+  parse paths gained the same guarantee — the content-type rejection, the
+  bounded body read, decompression and message decode all report the
+  metadata their response delivered. As before, the error's trailers exclude
+  `grpc-status`, `grpc-message` and `grpc-status-details-bin`, whose content
+  is already the error's code, message and details. That filter now also
+  applies on the Connect client-streaming path, so identical wire bytes give
+  identical `ConnectError::trailers()` whatever the call shape;
+  `ServerStream::trailers()` still reports the wire trailers verbatim.
+
+  [#202]: https://github.com/connectrpc/connect-rust/issues/202
+time: 2026-07-23T22:24:14.651900498-07:00
--- a/.changes/unreleased/Fixed-20260723-224245.yaml
+++ b/.changes/unreleased/Fixed-20260723-224245.yaml
@@ -1,0 +1,31 @@
+kind: Fixed
+body: |-
+  **A gateway propagating an upstream error could produce a response the
+  client could not read, losing the handler's error code** ([#202]). When a
+  handler returns a `ConnectError` that came back from a client call — the
+  ordinary gateway shape — that error carries the upstream response's
+  headers, and the server echoed them onto its own response verbatim.
+
+  Over Connect on HTTP/1.1 the consequence was total: the forwarded
+  `content-length` contradicted the body actually being written, hyper's
+  HTTP/1 encoder refused to serialize the response, and the connection
+  closed with zero bytes sent. Measured against connect-go v1.19.1, a
+  handler returning `permission_denied` surfaced at the caller as
+  `unavailable` with empty metadata, because a transport failure was all
+  the client ever saw; every handler error code collapsed the same way. A
+  forwarded `content-encoding: gzip` was a second, milder break — the
+  client tried to gunzip a plain-JSON error body, discarded it, and fell
+  back to the HTTP status. A forwarded `date` replaced the server's own,
+  since hyper emits no `Date` of its own once one is set, so responses
+  advertised the upstream's generation time.
+
+  The server now keeps its own value for any header it already set, which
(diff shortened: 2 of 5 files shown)


iainmcgin:
Filed the conformance gap upstream as connectrpc/conformance#1121: the whole testsuite only ever echoes `x-custom-header` and `x-custom-trailer`, which is why this shipped through 3600/0 server conformance.

The upstream issue puts the open question plainly — should a conformant server drop reserved names, forward them verbatim, or reject the RPC — and notes that `content-length` at least cannot be forwarded, since it yields a response the peer cannot read. The trailer half of the same question is #263 here.

asacamano (approved):

PR #264, opened 2026-08-01
envelope: decode without copying the body

allada:
Widen decode and decode_with_limit to impl Buf + Deref<Target = [u8]> and take the payload with copy_to_bytes, which Bytes and BytesMut both specialize to a refcount handoff. The gRPC unary, server-streaming and Connect client-streaming paths now decode straight out of the collected body instead of staging a BytesMut copy.

### DIFF
--- a/.changes/unreleased/Changed-20260801-120000.yaml
+++ b/.changes/unreleased/Changed-20260801-120000.yaml
@@ -1,0 +1,18 @@
+kind: Changed
+body: |-
+  **Envelope decoding no longer copies the request body.** `Envelope::decode`
+  and `Envelope::decode_with_limit` now accept any
+  `impl Buf + Deref<Target = [u8]>`, which both `Bytes` and `BytesMut` satisfy,
+  and take the payload with `Buf::copy_to_bytes`. Both types specialize that to
+  a refcount handoff rather than a memcpy, so a caller holding a fully-collected
+  body decodes straight out of it instead of staging a `BytesMut` copy first.
+  This removes a full-body copy per gRPC and gRPC-Web unary request, per
+  server-streaming request, and per Connect client-streaming response. The copy
+  was of the whole collected body, so on the server it was bounded by
+  `Limits::max_request_body_size` (4 MiB by default, and often raised), not by
+  `max_message_size`.
+
+  Decoded payloads now alias the buffer they were decoded from instead of an
+  exact-size copy, so retaining a request or response message keeps its whole
+  source buffer alive. Copy the fields you retain if that matters.
+time: 2026-08-01T12:00:00.000000000Z
--- a/connectrpc/src/client/mod.rs
+++ b/connectrpc/src/client/mod.rs
@@ -4012,7 +4012,7 @@ fn parse_connect_client_stream_envelopes(
     max_msg_size: usize,
     resp_headers: &http::HeaderMap,
 ) -> Result<(Bytes, http::HeaderMap), ConnectError> {
-    let mut buf = BytesMut::from(body.as_ref());
+    let mut buf = body;
     let mut message: Option<Bytes> = None;
     let mut trailers = http::HeaderMap::new();
     let mut saw_end_stream = false;
--- a/connectrpc/src/envelope.rs
+++ b/connectrpc/src/envelope.rs
@@ -145,7 +145,9 @@ impl Envelope {
     ///
     /// **Warning:** This method has no size limit. Use [`decode_with_limit`](Self::decode_with_limit)
     /// for untrusted input to prevent denial-of-service attacks.
-    pub fn decode(buf: &mut BytesMut) -> Result<Option<Self>, ConnectError> {
+    pub fn decode(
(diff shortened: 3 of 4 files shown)


iainmcgin (commented):
[claude code] Thanks for this — the core claim checks out against `bytes` 1.12.1: `Bytes::copy_to_bytes` is `split_to` (a move at the exact-fit boundary the server hits, no refcount traffic), `BytesMut::copy_to_bytes` is byte-for-byte the `split_to(len).freeze()` this replaces, and the three deleted `BytesMut::from(&body[..])` calls were the only full-body memcpys left in non-test code. Rebased onto `main` it passes the full CI-equivalent gate locally (one trivial conflict in `parse_connect_client_stream_envelopes`, which `main` split into a `scan_*` helper — the one-line change moves into the helper).

A few things to change before this can merge:

**1. Call the signature change out as breaking.** Widening `&mut BytesMut` to `&mut (impl Buf + Deref<Target = [u8]>)` compiles for every caller passing `&mut buf` where `buf: BytesMut` or `Bytes`, but it breaks two real call shapes: deref-coerced arguments (`Envelope::decode(&mut guard)` with `guard: MutexGuard<'_, BytesMut>` now fails with `MutexGuard: Deref<Target = [u8]>` not satisfied, since coercion never fires against a generic parameter), and locals whose type was pinned by the parameter (`let mut buf = body.into(); Envelope::decode(&mut buf)` is now E0282). The changelog uses a scoped marker for this — e.g. `**Breaking, for direct `Envelope::decode` callers only.**` — plus one migration sentence (`&mut *guard`, or annotate the local).

**2. Prefer a named generic over `impl Trait`.** `pub fn decode<B: Buf + Deref<Target = [u8]> + ?Sized>(buf: &mut B)` is semver-equivalent, renders as an ordinary generic in rustdoc, and lets a caller hit by (1) write `Envelope::decode::<BytesMut>(&mut buf)`; with APIT that turbofish is a hard error. This crate's existing APIT is all by-value `impl Into<_>` sugar.

**3. Document the aliasing contract on the methods, and fix the changelog's before-state.** The fragment says payloads "now alias the buffer they were decoded from instead of an exact-size copy", but `BytesMut::split_to` + `freeze` already handed back a `Bytes` over the whole staging allocation, so retaining a message kept a body-sized buffer alive before this change too. What is actually new is *which* buffer: for a body that arrives in one frame, `Collected::to_bytes` returns the transport's frame verbatim, so the decoded message now aliases hyper's read buffer (h2 codec buffer, or the h1 adaptive buffer, which can be several hundred KiB). That belongs in the `///` on both methods — which types satisfy the bound, that `Bytes`/`BytesMut` hand off by refcount while `&[u8]` takes the default `copy_to_bytes` and copies, and that the payload aliases the source buffer, so copy what you retain. The changelog paragraph can then just point at that.

**4. State the `Deref`/`Buf` invariant.** The body reads the header through the `Deref` slice (`buf.len()`, `buf[0..5]`) and takes the payload through the `Buf` cursor (`advance`, `copy_to_bytes`). Nothing in the bound requires `deref().len() == remaining()`; a `Buf` whose two views disagree either misparses silently or panics in `advance`/`copy_to_bytes`. No type in std or `bytes` does that today, but this is a public fn: document the invariant, add `debug_assert_eq!(buf.remaining(), buf.len())` at the top of `decode_with_limit`, and a `# Panics` section.

Smaller:

- The changelog's "4 MiB by default, and often raised" — both `max_request_body_size` and `max_message_size` default to the same 4 MiB, so the contrast is empty at defaults, and "often raised" is a claim about deployments we can't back. "Bounded by `max_request_body_size` rather than `max_message_size`, so it grew with whatever the body limit was raised to" says the same thing without it.
- The new `buf.len() == HEADER_SIZE` assertion on the oversized-frame path is only proving the length didn't drop to zero, because the fixture is exactly five header bytes. Append a complete second envelope and assert a follow-up `decode_with_limit` still returns it intact — that's the desync the comment names.
- `// Avoid a copy of our data if possible.` — say what makes it possible: `Bytes` and `BytesMut` specialise `copy_to_bytes` to a refcount split.
- The test section header and the doc comments on the new tests argue with a hypothetical reviewer ("a cross-comparison would still pass") rather than describing the test; the sibling headers use the `// ── ... ──` style.
- The fragment timestamp is hand-typed; `task changelog-new` generates one.

Iain will do the final review and merge; happy to push the rebase onto your branch if you'd rather, just say.

iainmcgin:
[claude code] Thanks for this, and for the careful analysis of where the body copies were. #300 has since removed the same three copies (client Connect stream parsing, gRPC unary, and server streaming): it added `Envelope::decode_bytes_with_limit(&mut Bytes)`, which takes the payload as a zero-copy slice, and kept `decode` / `decode_with_limit` on `&mut BytesMut`, so the change landed without widening a public signature. That leaves nothing here that `main` still needs, and the branch now conflicts in `envelope.rs`, `service.rs` and `client/mod.rs`, so I'm closing this as superseded by #300. If you see a copy on the decode path that #300 missed, a fresh PR against `main` would be very welcome.

PR #265, opened 2026-08-02
client: return one-frame response bodies without copying

allada:
collect_body_bounded copied every frame into a BytesMut, including the common case of a body that arrives whole in one frame. Hold the first frame by reference count and return it directly when no second frame follows; promote to a buffer only when one does.

Using http_body_util::Limited here instead would need B::Error to be std::error::Error rather than Display, which propagates to ServerStream and BidiStream and the transport trait's body. Not worth a public API break for this.

### DIFF
--- a/.changes/unreleased/Changed-20260801-190000.yaml
+++ b/.changes/unreleased/Changed-20260801-190000.yaml
@@ -1,0 +1,9 @@
+kind: Changed
+body: |-
+  **Client response bodies that arrive in one frame are no longer copied.**
+  `collect_body_bounded` holds the first data frame by reference count and
+  returns it directly when no second frame follows, which is the common case
+  for unary responses and error bodies. Multi-frame bodies concatenate as
+  before, now starting from an exactly-sized buffer. The size limit is still
+  enforced before any frame is retained.
+time: 2026-08-01T19:00:00.000000000Z
--- a/connectrpc/src/client/mod.rs
+++ b/connectrpc/src/client/mod.rs
@@ -4368,22 +4368,42 @@ where
     B: Body<Data = Bytes>,
     B::Error: std::fmt::Display,
 {
-    let mut buf = BytesMut::new();
     let mut stream = std::pin::pin!(body);
+    // Hold the first frame by refcount so a one-frame body is never copied.
+    let mut head: Option<Bytes> = None;
+    let mut joined: Option<BytesMut> = None;
+    let mut len: usize = 0;
+
     loop {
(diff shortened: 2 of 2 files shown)


iainmcgin (commented):
[claude code] Thanks — the rewrite is correct on the points that matter: the running `len` equals the old `buf.len()` at every check (empty frames add zero), the `> max_size` boundary is unchanged so a body of exactly `max_size` still succeeds, nothing is retained before the limit passes, and trailer frames behave as before. `Full` hands out its stored `Bytes` unchanged, so the pointer-equality test is valid and can't false-pass while `src` is alive.

Two things before it can merge:

**1. It no longer compiles after a rebase onto `main`.** `main` gave `collect_body_bounded` a third `deadline: Option<Instant>` argument (post-deadline body errors are classified as `deadline_exceeded`, #238), and the merge is textually clean because the hunks don't touch that arm — but the two new tests call the two-argument form and fail with E0061. `, None` on both fixes it locally; the rest of the gate then passes. Please also keep the `classify_body_read_error` arm intact when you resolve, since taking the PR side of that region would silently revert it.

**2. Say that the returned `Bytes` may alias the transport buffer.** For a one-frame body the value handed back is hyper's frame slice — a piece of the h2 codec buffer, or of the h1 adaptive read buffer, which grows to several hundred KiB — and the Connect unary path keeps it for the lifetime of the decoded response. That matches what `http_body_util::Collected::to_bytes` does, and we're going with zero-copy by default rather than a size threshold or a config option; the caller-local remedy is to copy what you retain. So it needs to be in the doc comment on `collect_body_bounded`, and the changelog should say it rather than naming `collect_body_bounded` itself, which is private and not something a user can look up. (The changelog's "exactly-sized buffer" is also only exact for a two-frame body; a third frame reallocates. "Sized for the frames seen so far", or drop the clause.)

Smaller:

- The `data.is_empty() { continue }` branch has no test. With `ChannelBody`, frames `["", "foo"]` should return `"foo"` by refcount, and `["foo", ""]` should not promote to a `BytesMut`.
- The multi-frame test's non-aliasing assertion compares against a `from_static` pointer in rodata, which a heap buffer can never equal, so it doesn't catch anything the `b"foobarbaz"` equality doesn't already; fine to drop.
- The final `match` relies on `head` being `None` whenever `joined` is `Some` (true, via `head.take()`); a `debug_assert!` pins it.
- "nothing else pins it" in the test doc isn't quite right — `src` holds the second refcount, which is what makes the assertion safe.
- Fragment timestamp is hand-typed; `task changelog-new` generates one.

This stacks cleanly on #264. Iain will do the final review and merge.

PR #269, opened 2026-08-13
interceptor: plumbing for re-runnable chains and lazily-encoded payloads

aknott-ant:
First of a short stack adding client-side interceptors on the existing `Interceptor` trait. This PR is additive plumbing on the shared types; nothing on the dispatch path changes.

**Adds**

- `RequestContext::headers_mut()` so an interceptor can rewrite request headers (auth, trace propagation).
- `Payload::from_message(msg, format)`: a payload built from a typed message, encoded lazily (and once) on `encoded()`. Used for client requests later in the stack, and by interceptors that short-circuit with a synthesized response.
- `Payload::try_clone()`, `UnaryRequest::{from_parts, try_clone}`, and `Next: Clone`, which together let an interceptor run the rest of the chain more than once:
  ```rust
  let spare = req.try_clone()?;
  match next.clone().run(req).await {
      Err(e) if e.code == ErrorCode::Unavailable => next.run(spare).await,
      done => done,
  }
  ```
- Internal `InterceptorChain` newtype for the chain storage (`ConnectRpcService` now, `ClientConfig` next).

**Decisions**

- `Next` is `Clone` rather than `Copy`, so a re-run is an explicit `next.clone()`. `NextStream` stays non-`Clone` since an inbound stream can't be replayed.
- `Payload` stays non-`Clone`; `try_clone` is the explicit, wire-equivalent copy with a fresh decode cache.

Unit tests for each addition; lint, tests, and docs clean.

### DIFF
--- a/.changes/unreleased/Added-20260813-043124.yaml
+++ b/.changes/unreleased/Added-20260813-043124.yaml
@@ -21,8 +21,8 @@ body: |-
   - **`UnaryRequest::from_parts(ctx, payload)`** — constructor for the
     `#[non_exhaustive]` request from an existing `Payload`; `StreamRequest`
     is now `Clone`.
-  - **`Next` and `NextStream` are now `Clone`**, so an interceptor can run
-    the rest of the chain more than once (`next.clone().run(first)` then
+  - **`Next` is now `Clone`**, so a unary interceptor can run the rest of
+    the chain more than once (`next.clone().run(first)` then
     `next.run(second)`); the consuming `run(self)` keeps the single-call
     case explicit.
 time: 2026-08-13T04:31:36.268465697Z
--- a/.changes/unreleased/Added-20260813-043124.yaml
+++ b/.changes/unreleased/Added-20260813-043124.yaml
@@ -1,0 +1,28 @@
+kind: Added
+body: |-
+  Interceptor plumbing that a re-runnable (retry-style) chain and the upcoming
+  client-side interceptors both need. All additive; no dispatch behavior
+  changes:
+
+  - **`RequestContext::headers_mut()`** — mutable access to request headers,
+    the accessor an interceptor uses to inject auth or trace metadata before
+    `next.run(req)`. (`extensions_mut()` already existed; the docs claimed
+    headers were mutable but no public accessor allowed it.)
+  - **`Payload::from_message(msg, format)`** — the lazily-*encoded*
+    counterpart of `Payload::new`: wraps a typed message and encodes only when
+    `encoded()` runs, memoizing the result (a replacement set with
+    `set_message` is now also encoded at most once). Lets an interceptor
+    short-circuit with a synthesized response, and is how the client call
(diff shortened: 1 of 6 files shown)


iainmcgin:
[claude code] Reviewed for 0.9.0 — two passes (correctness and downstream-API), both approve. The memo is sound (`OnceLock` in a boxed `Replacement`, a new replacement is a new memo so it can never be stale, a failed encode is not cached), `Next: Clone` is two pointer copies over a slice walk with no consumed state so two clones can each `run` once, `headers_mut` cannot stale anything because every protocol-derived value is captured in locals before the context is built (checked `RequestMetadata::from_headers`, the deadline, codec, and the response-compression negotiation at `service.rs:2042`), and `InterceptorChain` is behaviour-preserving. The retry test asserting the *inner* interceptor ran twice is the assertion such tests usually get wrong — nice.

You had already restacked onto current `main` by the time I pushed, so the only change to your branch is one maintainer commit on top (your commits are untouched):

**The negotiated format wins.** `Payload::from_message(msg, format)` leaves `format` to the caller, and `into_encoded()` encoded with the payload's own format — so an interceptor that short-circuits with the obvious `from_message(reply, CodecFormat::Proto)` on a JSON call sent proto bytes under `application/json`. The guide's example did the right thing (`req.payload.format()`), but the wrong spelling is shorter. There's now `Payload::encoded_as(format)` — a message payload is simply encoded in the requested format; wire bytes in another format are an `internal` error rather than a body the peer can't parse — and `call_unary_intercepted` uses it, so the peer always gets what it negotiated. Test drives a proto-built short-circuit through a JSON call and decodes the result as JSON.

**`try_clone` keeps a replacement body unbudgeted.** It copied `decode_options` onto bytes that were encoded from a locally-held message, which `view()` deliberately exempts from the wire limits (#258); a retry attempt could then fail `invalid_argument` for a message the server itself built. The copy now keeps buffa's defaults in that case. Also the memo returns the winning allocation to a racing loser (`get_or_init`).

**Docs:** `headers_mut` now says its "changes only what downstream reads" claim is the server-side chain — on a client chain the edited map *is* the wire, which is exactly the auth/trace use case; `NextStream` says streaming chains run once by construction; `Next` notes a retry shares the request deadline and span (open a span per attempt for OTel); `Payload::format` covers the `from_message` case; the fragment no longer refers to unshipped work.

Flagged, not changed — your call: `UnaryRequest::from_parts` drops the context's decode limits (documented; a `from_parts_with_limits` or stamping when the payload is still at defaults would close it); a tracing interceptor has no cheap message-size accessor (`bytes()` is empty for `from_message`, `encoded()` forces the encode — a `Payload::wire_len()` would fit); and the `from_message(msg, format)` signature itself, given the format is now reconciled at the boundary anyway.

665 lib tests + minimal-features 490, clippy and rustdoc `-Dwarnings` clean.

iainmcgin (approved):

PR #270, opened 2026-08-13
client: identify the RPC by Spec in call_*; clients pass FOO_SPEC.with_origin(Client)

aknott-ant:
> Stacked on #269 (first two commits, already rebased onto current `main`); will rebase again when it lands.

Second step toward client-side interceptors: the client runtime now receives the method's `Spec` (procedure, stream type, idempotency, origin) instead of two strings, so a client interceptor can see what a server interceptor sees.

**Changes**

- **Breaking:** `connectrpc::client::call_unary` / `call_unary_get` / `call_server_stream` / `call_client_stream` / `call_bidi_stream` take `spec: Spec` in place of `service: &str, method: &str`. Generated clients pass `FOO_SERVICE_BAR_SPEC.with_origin(SpecOrigin::Client)`; hand-written callers do the same or use `Spec::client("/pkg.Svc/Method", StreamType::…)`. Wire paths are unchanged.
- New `Spec::with_origin(SpecOrigin)` and `Spec::same_method(&Spec)` (the client-side value is not `==` to the constant, since `origin` differs).
- A spec with the wrong stream shape or a server origin is rejected with `Internal` before anything is sent, with a message naming the fix.
- Request URIs are built structurally from the base URI's parsed parts, in one helper instead of five `format!` + re-parse copies.
- Conformance client updated; six checked-in generated dirs regenerated.

**Decisions**

- Replace the string pair rather than add parallel `_spec` entry points: generated code is the real caller and the crate is pre-1.0. A dynamic caller interns its `&'static` procedure (documented on `Spec::client`).
- One `Spec` constant per method plus `with_origin`, rather than a second `*_CLIENT_SPEC` constant per method: origin is a call-site property (as in connect-go), and this avoids doubling generated constants and docs.

Client conformance (Connect, gRPC, gRPC-Web) passes; wire behavior unchanged.

### DIFF
--- a/.changes/unreleased/Added-20260813-043124.yaml
+++ b/.changes/unreleased/Added-20260813-043124.yaml
@@ -21,8 +21,8 @@ body: |-
   - **`UnaryRequest::from_parts(ctx, payload)`** — constructor for the
     `#[non_exhaustive]` request from an existing `Payload`; `StreamRequest`
     is now `Clone`.
-  - **`Next` and `NextStream` are now `Clone`**, so an interceptor can run
-    the rest of the chain more than once (`next.clone().run(first)` then
+  - **`Next` is now `Clone`**, so a unary interceptor can run the rest of
+    the chain more than once (`next.clone().run(first)` then
     `next.run(second)`); the consuming `run(self)` keeps the single-call
     case explicit.
 time: 2026-08-13T04:31:36.268465697Z
--- a/.changes/unreleased/Added-20260813-043124.yaml
+++ b/.changes/unreleased/Added-20260813-043124.yaml
@@ -1,0 +1,28 @@
+kind: Added
+body: |-
+  Interceptor plumbing that a re-runnable (retry-style) chain and the upcoming
+  client-side interceptors both need. All additive; no dispatch behavior
+  changes:
+
+  - **`RequestContext::headers_mut()`** — mutable access to request headers,
+    the accessor an interceptor uses to inject auth or trace metadata before
+    `next.run(req)`. (`extensions_mut()` already existed; the docs claimed
+    headers were mutable but no public accessor allowed it.)
+  - **`Payload::from_message(msg, format)`** — the lazily-*encoded*
+    counterpart of `Payload::new`: wraps a typed message and encodes only when
+    `encoded()` runs, memoizing the result (a replacement set with
+    `set_message` is now also encoded at most once). Lets an interceptor
+    short-circuit with a synthesized response, and is how the client call
+    path will hand typed requests to the chain. `bytes()` is empty for such a
+    payload; `Payload`'s `Debug` output now labels that field `wire_len`.
+  - **`Payload::try_clone()`** and **`UnaryRequest::try_clone()`** — explicit
+    copies (encoded bytes, cloned context, empty decode cache) for an
+    interceptor that needs a second request.
+  - **`UnaryRequest::from_parts(ctx, payload)`** — constructor for the
+    `#[non_exhaustive]` request from an existing `Payload`; `StreamRequest`
+    is now `Clone`.
+  - **`Next` and `NextStream` are now `Clone`**, so an interceptor can run
+    the rest of the chain more than once (`next.clone().run(first)` then
+    `next.run(second)`); the consuming `run(self)` keeps the single-call
+    case explicit.
+time: 2026-08-13T04:31:36.268465697Z
--- a/.changes/unreleased/Added-20260813-043124.yaml
+++ b/.changes/unreleased/Added-20260813-043124.yaml
@@ -1,7 +1,7 @@
 kind: Added
 body: |-
-  Interceptor plumbing that a re-runnable (retry-style) chain and the upcoming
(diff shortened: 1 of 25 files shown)


iainmcgin:
[claude code] Reviewed for 0.9.0 — two passes, both approve with one required fix, now applied. The `Spec`-threading matches connect-go's `Spec{IsClient: true}`, the five entry points validate before any transport touch, and the structural URI builder is byte-identical to the old `format!` for every base shape I could construct (trailing slash, path prefix, IPv6 authority, GET query) while fixing the `"/?a=1/pkg.Svc/Do"` corruption a base query used to cause. Client conformance (Connect, gRPC, gRPC-Web) passes on the rebased stack, and `task generate:all` under CI's exact pins (buf 1.69.0, protoc 33.5, buffa v0.9.1) reproduces your regenerated trees byte-for-byte.

You had already restacked onto current `main` by the time I pushed, so your branch gains two maintainer commits and nothing else: the #269 fixup it now sits on, and this one:

**Panic → error.** `procedure_uri`'s no-prefix fast path used `PathAndQuery::from_static`, which is `from_shared(..).unwrap()`: a space, `#`, or any byte ≥ 0x80 in a procedure panicked in release, while the same `Spec` against `http://h/api` returned `internal`. `check_client_spec` only checks the slash shape, so this was reachable from exactly the interning dynamic caller the docs describe. Both branches now go through `from_maybe_shared(Bytes::from_static(..))` — still no allocation — and report `internal`.

**A codegen rule that rejected valid protos.** `check_method_collisions` treated `Get` + `GetSpec` as a clash on `get_spec`, but the constants are `X_GET_SPEC` and `X_GET_SPEC_SPEC`, and `get_spec` is only ever a client method name. That pair compiled on 0.8 and now does again; the test pins it. (The cross-service case — service `XGet` method `Spec` vs service `X` method `Get` both emitting `X_GET_SPEC` — is unchecked before and after; worth a follow-up.)

**Docs.** `Spec`'s type doc said it could be "compared freely"; it now says `PartialEq`/`Hash` include `origin`, which is why a client-side value is not `==` the constant and a `HashMap` keyed by constants misses it, and points at `same_method` — which takes its argument by value since `Spec: Copy`. `# Panics` on `Spec::client`/`server`. The per-constant doc the generator emits is cut from a paragraph to one sentence (−217 lines across the checked-in trees; a forty-method service was going to carry forty copies). `ClientConfig::new` documents how request URIs are formed and that a base query is dropped; the changelog snippet gains its imports.

**Conformance client fidelity.** It used the generated constant regardless of the `service`/`method` the suite named; `client_spec()` keeps the constant when they match and builds an ad-hoc client `Spec` otherwise, so a future suite case is never silently re-addressed.

Two design points both reviewers raised that I've left as you had them — say if you'd like either changed before merge: (1) the `origin != Client` check errors at runtime for the natural `call_unary(&t, &cfg, GREET_SPEC, ..)`, where normalising with `with_origin` internally (or a `ClientSpec` newtype) would make it unrepresentable; strict-now can be relaxed later without a break, so shipping strict is defensible. (2) `Internal` for the wrong-shape / wrong-origin spec vs `invalid_argument`, which `call_unary_get` uses for the adjacent "non-Connect protocol" caller error ten lines away — connect-go uses `CodeInternal` for these, so either is principled, but a sentence on why would stop it reading as an oversight.

910 workspace tests, clippy and rustdoc `-Dwarnings` clean.

aknott-ant:
Thanks for carrying these into the 0.9.0 cut, and for the two fixes — both are right. Pushing onto the fork branches is fine.

On the two design points, keeping both as they are:

1. **Strict origin check.** A server-origin `Spec` reaching a client entry point is a caller bug, and quietly normalising it would hide exactly the mistake the check exists to catch (an interceptor later observing `origin: Server` on a client call). Since strict can be relaxed without a break, I'd rather ship strict and see whether anyone actually hits it.
2. **`Internal` rather than `invalid_argument`.** The distinction I'm drawing: `invalid_argument` in `call_unary_get` is about the *configuration* (a Connect-only feature on a non-Connect client), which a user can hit with valid code; the spec checks are *programming errors* that generated code cannot produce and that mean the caller's code is wrong. That matches `ConnectError::internal`'s use elsewhere in the crate for caller bugs, and connect-go's `CodeInternal`. The rustdoc on `check_client_spec` says so ("all are caller bugs, so the error is `Internal`"); happy to add the sentence to each entry point's `# Errors` if you'd prefer it there.

One small thing I found running an end-to-end harness against your commit, not blocking: `http`'s `PathAndQuery` treats `#` as a fragment delimiter, so a hand-built `Spec::client("/pkg.Svc/Get#Order", ..)` parses as path `/pkg.Svc/Get` and the call goes out to the wrong method (server answers `Unimplemented`) instead of being rejected; non-ASCII is accepted as UTF-8 and goes out too. Only reachable from hand-built specs, and `procedure_is_well_formed` is the natural place for a `#`/`?` check if you want it in the cut; otherwise I'll fold it into the next PR.

iainmcgin (approved):

PR #296, opened 2026-09-06
perf: trim per-request allocations on the unary paths

iainmcgin:
Profiling the small-unary gRPC benchmark (`cross/unary_small_grpc`, the ~300-byte `small_request()` payload) showed the server making 63 heap allocations per request, about 26 of them in connect-rust's own request path rather than in decode or the handler. This removes nine of them and one full copy of every request body:

- `handle_grpc_unary_request` (and the server-streaming path) copied the collected body into a `BytesMut` to parse the 5-byte envelope. `Envelope::decode_from_bytes_with_limit` splits the payload out of the `Bytes` by reference count.
- `RequestMetadata::from_headers` cloned the request `HeaderMap` for the handler context although the request is consumed a few lines later. The three entry points now take the request apart first and move the map.
- Constant header values (content-type, encodings, `te`, `connect-protocol-version`) on server responses and client requests use `HeaderValue::from_static`, and the accept-encoding value is built once per `CompressionRegistry` (now exposed as `accept_encoding_value()`), instead of each going through `HeaderValue::from_str`.

`from_static` on a negotiated encoding name is only safe if provider names are valid header values, so `CompressionRegistry::register` now panics when `CompressionProvider::name()` is not an RFC 9110 token. The alternative was caching a `HeaderValue` per provider and keeping `register` infallible.

Measured locally with a counting allocator and `perf stat`, single worker, closed loop: 63 → 54 allocations and 87k → 84k instructions per request on the server. Latency at c=1 is loopback-bound and does not move.

### DIFF
--- a/.changes/unreleased/Added-20260906-162100.yaml
+++ b/.changes/unreleased/Added-20260906-162100.yaml
@@ -1,0 +1,7 @@
+kind: Added
+body: |-
+    `Envelope::decode_from_bytes_with_limit` unframes an envelope from a
+    collected `Bytes` body without first copying it into a `BytesMut`, and
+    `CompressionRegistry::accept_encoding_value` exposes the registry's
+    supported-encodings list as a ready-made `HeaderValue`.
+time: 2026-09-06T16:21:00.000000000+00:00
--- a/.changes/unreleased/Changed-20260906-155000.yaml
+++ b/.changes/unreleased/Changed-20260906-155000.yaml
@@ -1,0 +1,11 @@
+kind: Changed
+body: |-
+    **Fewer allocations per request.** A unary gRPC request is unframed out of
+    the collected body by reference count instead of being copied first (one
+    fewer full copy of every request body), the request's header map is moved
+    into the handler's `RequestContext` rather than cloned, and the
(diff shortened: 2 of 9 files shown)


iainmcgin:
[claude code] Bare-metal A/B for this change (AWS c7i.metal-24xl, Xeon Platinum 8488C, turbo off; server pinned to 8 cores, load generator on 16 others, loopback). `bench_server` built from `main` and from this branch, driven by the same closed-loop gRPC small-unary client (`unary_load`, `HttpClient` over h2), 10 s per cell, three rounds alternating the two servers:

| concurrency | main | this branch | delta |
|---:|---:|---:|---:|
| 1 | 68.0 µs/req | 67.3 µs/req | −1.1% |
| 16 | 39.7k req/s | 40.2k req/s | +1.3% |
| 64 | 40.1k req/s | 40.4k req/s | +0.9% |

The branch was ahead in all nine rounds; round-to-round spread was about ±1%. Small, as expected for a path that is dominated by the kernel round trip at c=1 and by the single connection's h2 lock at c=16/64; the allocation count (63 → 54 per request) and instruction count (about −4%) are the more direct measures of what changed.

rpb-ant (approved):

PR #297, opened 2026-09-06
client: share an Http2Connection without a buffer worker task

### LINKED ISSUE #285: Http2Connection: first request after a graceful GOAWAY fails with `h2 send failed: operation was canceled` while a stream is open
When a server sends a graceful HTTP/2 GOAWAY on a connection that still has a stream open, the next request routed to that connection through `Http2Connection` (used via `SharedHttp2Connection`, which the client module docs list as the default transport for gRPC) fails with `unavailable: h2 send failed: operation was canceled`. The request is never written to the wire. The request after it triggers the reconnect and succeeds. A graceful GOAWAY is the frame a server sends to stop accepting new streams on a connection while letting open streams finish; connectrpc's own server sends one from `with_max_connection_age`, `with_max_connection_idle` and `with_max_requests_per_connection`, so a client that keeps a long-lived stream open against such a server fails at least one call on each recycle that happens while the stream is open.

The precondition is exactly "a stream is still open when the GOAWAY arrives". An idle connection closes fully, `poll_ready` reports it, and `Reconnect` re-dials before any request is lost. `HttpClient` (the hyper-util pooled transport) is not affected in the same scenario because hyper-util re-sends the unstarted request.

## Versions

connectrpc 0.9.0, features `client` + `server` (same behaviour on 0.8.1). The program below resolved hyper 1.11.0, hyper-util 0.1.20, h2 0.4.16, tokio 1.53.1.

## Reproduction

One binary, no codegen. The server is connectrpc's own `Server` with `with_max_connection_age(1.5 s)` and a 120 s grace, serving a server-streaming method that yields one message and then never ends, plus a unary echo. Each scenario opens the stream (or not), makes one unary call, waits past the connection age, then makes three more unary calls on the same transport.

<details>
<summary>Repro program (Cargo.toml + src/main.rs)</summary>

`Cargo.toml`:

```toml
[package]
name = "goaway-repro"
version = "0.1.0"
edition = "2024"
publish = false

[dependencies]
connectrpc = { version = "=0.9.0", features = ["client", "server"] }
buffa-types = { version = "0.9.1", features = ["json"] }
tokio = { version = "1", features = ["full"] }
futures = "0.3"
http = "1"
bytes = "1"
tracing-subscriber = { version = "0.3", features = ["env-filter", "fmt"] }

[workspace]
```

`src/main.rs`:

```rust
use std::time::{Duration, Instant};

use buffa_types::google::protobuf::StringValue;
use buffa_types::google::protobuf::__buffa::view::StringValueView;
use connectrpc::client::{
    call_server_stream, call_unary, CallOptions, ClientConfig, ClientTransport, Http2Connection,
    HttpClient,
};
use connectrpc::server::Server;
use connectrpc::{handler_fn, streaming_handler_fn, Protocol, Response, Router, ServiceStream, Spec, StreamType};
use futures::StreamExt;

const HANG: Spec = Spec::client("/test.v1.T/Hang", StreamType::ServerStream);
const PING: Spec = Spec::client("/test.v1.T/Ping", StreamType::Unary);
const MAX_AGE: Duration = Duration::from_millis(1500);

fn router() -> Router {
    Router::new()
        // A server stream that yields one message and then stays open.
        .route_server_stream(
            "test.v1.T",
            "Hang",
            streaming_handler_fn(|_ctx, _req: StringValue| async move {
                let s = futures::stream::once(async { Ok(StringValue::from("first")) })
                    .chain(futures::stream::pending());
                Ok(Response::new(Box::pin(s) as ServiceStream<StringValue>))
            }),
        )
        .route("test.v1.T", "Ping", handler_fn(|_ctx, req: StringValue| async move { Ok(Response::new(req)) }))
}

/// Server that sends a graceful GOAWAY at MAX_AGE and keeps open streams for 120 s.
async fn serve() -> http::Uri {
    let bound = Server::bind("127.0.0.1:0")
        .await
        .unwrap()
        .with_max_connection_age(MAX_AGE)
        .with_max_connection_age_grace(Duration::from_secs(120));
    let addr = bound.local_addr().unwrap();
    tokio::spawn(async move { bound.serve(router()).await.unwrap() });
    format!("http://{addr}").parse().unwrap()
}

async fn scenario<T>(label: &str, transport: &T, uri: http::Uri, hold_stream: bool)
where
    T: ClientTransport,
    T::ResponseBody: connectrpc::http_body::Body<Data = bytes::Bytes> + Send + Unpin,
    <T::ResponseBody as connectrpc::http_body::Body>::Error: std::fmt::Display,
{
    let cfg = ClientConfig::new(uri).with_protocol(Protocol::Grpc);
    let mut held = None;
    if hold_stream {
        let mut s = call_server_stream::<_, StringValue, StringValueView<'static>>(
            transport, &cfg, HANG, StringValue::from("hang"), CallOptions::default(),
        )
        .await
        .expect("open Hang");
        s.message::<StringValue>().await.expect("first message").expect("some");
        println!("{label}: stream open, first message received");
        held = Some(s);
    }
    let ping = || async {
        let t = Instant::now();
        let r = call_unary::<_, StringValue, StringValueView<'static>>(
            transport, &cfg, PING, StringValue::from("ping"), CallOptions::default(),
        )
        .await;
        match r {
            Ok(_) => format!("Ok (took {:?})", t.elapsed()),
            Err(e) => format!(
                "Err code={:?} message={:?} source={:?} (took {:?})",
                e.code, e.message, std::error::Error::source(&e).map(|s| s.to_string()), t.elapsed()
            ),
        }
    };
    println!("{label}: ping before GOAWAY: {}", ping().await);
    tokio::time::sleep(MAX_AGE * 2 + Duration::from_millis(500)).await; // past max_connection_age (+-10% jitter)
    for i in 1..=3 {
        println!("{label}: ping #{i} after GOAWAY: {}", ping().await);
    }
    if let Some(mut s) = held {
        let state = match tokio::time::timeout(Duration::from_millis(300), s.message::<StringValue>()).await {
            Err(_) => "still open (pending)".to_owned(),
            Ok(Ok(None)) => "ended cleanly".to_owned(),
            Ok(Ok(Some(_))) => "unexpected message".to_owned(),
            Ok(Err(e)) => format!("errored: {:?} {:?}", e.code, e.message),
        };
        println!("{label}: held stream afterwards: {state}");
    }
}

#[tokio::main]
async fn main() {
    // Optional: RUST_LOG=h2=debug,hyper_util=trace,connectrpc=debug for a frame trace on stderr.
    tracing_subscriber::fmt()
        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
        .with_writer(std::io::stderr)
        .with_ansi(false)
        .init();
    println!("--- A: SharedHttp2Connection, a server stream is open when GOAWAY arrives");
    let uri = serve().await;
    let conn = Http2Connection::connect_plaintext(uri.clone()).await.unwrap().shared(64);
    scenario("A", &conn, uri, true).await;

    println!("--- B: HttpClient::plaintext_http2_only(), same scenario (contrast)");
    let uri = serve().await;
    scenario("B", &HttpClient::plaintext_http2_only(), uri, true).await;

    println!("--- C: SharedHttp2Connection, connection idle when GOAWAY arrives (control)");
    let uri = serve().await;
    let conn = Http2Connection::connect_plaintext(uri.clone()).await.unwrap().shared(64);
    scenario("C", &conn, uri, false).await;
}
```

</details>

<details>
<summary>Observed output (<code>cargo run</code>), scenarios A, B, C</summary>

Observed (`cargo run`):

```
--- A: SharedHttp2Connection, a server stream is open when GOAWAY arrives
A: stream open, first message received
A: ping before GOAWAY: Ok (took 1.439839ms)
A: ping #1 after GOAWAY: Err code=Unavailable message=Some("h2 send failed: operation was canceled") source=Some("operation was canceled") (took 375.83µs)
A: ping #2 after GOAWAY: Ok (took 3.442559ms)
A: ping #3 after GOAWAY: Ok (took 1.3978ms)
A: held stream afterwards: still open (pending)
--- B: HttpClient::plaintext_http2_only(), same scenario (contrast)
B: stream open, first message received
B: ping before GOAWAY: Ok (took 1.54378ms)
B: ping #1 after GOAWAY: Ok (took 3.84195ms)
B: ping #2 after GOAWAY: Ok (took 1.31747ms)
B: ping #3 after GOAWAY: Ok (took 1.35007ms)
B: held stream afterwards: still open (pending)
--- C: SharedHttp2Connection, connection idle when GOAWAY arrives (control)
C: ping before GOAWAY: Ok (took 1.99318ms)
C: ping #1 after GOAWAY: Ok (took 4.519769ms)
C: ping #2 after GOAWAY: Ok (took 1.42356ms)
C: ping #3 after GOAWAY: Ok (took 1.407ms)
```

</details>

Scenario A is the bug: ping #1 after the GOAWAY fails in well under a millisecond (no round trip happened) and the held stream is unharmed. Scenario B is the same sequence through `HttpClient::plaintext_http2_only()`: nothing fails, and with `RUST_LOG=hyper_util=trace` hyper-util logs the retry that makes the difference:

```
TRACE hyper_util::client::legacy::client: unstarted request canceled, trying again (reason=hyper_util::client::legacy::Error(Canceled, hyper::Error(Canceled, "connection closed")))
```

Scenario C is the control: with no stream open when the GOAWAY arrives, `SharedHttp2Connection` loses nothing.

<details>
<summary>h2 frame trace for scenario A</summary>

The h2 frame trace for scenario A (`RUST_LOG=h2=debug,connectrpc=debug`; abridged: date and level prefix, PING and DATA frames, the response's initial HEADERS, and the server's view of stream 3 omitted) shows the unary call before the GOAWAY on stream 3, the server's two-step GOAWAY while stream 1 is still open, and then no HEADERS frame at all for ping #1. The `reconnecting` line is logged from ping #2's `poll_ready`, after ping #1 has already failed:

```
16:02:56.829439 Connection{peer=Client}: h2::codec::framed_write: send frame=Headers { stream_id: StreamId(3), flags: (0x4: END_HEADERS) }
16:02:56.830416 Connection{peer=Client}: h2::codec::framed_read: received frame=Headers { stream_id: StreamId(3), flags: (0x5: END_HEADERS | END_STREAM) }
16:02:58.248571 Connection{peer=Server}: h2::codec::framed_write: send frame=GoAway { error_code: NO_ERROR, last_stream_id: StreamId(2147483647) }
16:02:58.248937 Connection{peer=Client}: h2::codec::framed_read: received frame=GoAway { error_code: NO_ERROR, last_stream_id: StreamId(2147483647) }
16:02:58.249233 Connection{peer=Server}: h2::codec::framed_write: send frame=GoAway { error_code: NO_ERROR, last_stream_id: StreamId(3) }
16:02:58.249384 Connection{peer=Client}: h2::codec::framed_read: received frame=GoAway { error_code: NO_ERROR, last_stream_id: StreamId(3) }
16:03:00.332382 connectrpc::client::http2: h2 connection lost; reconnecting
```

</details>

The error text is itself evidence that the request never left the process: hyper produces `Canceled` only for a request that was not yet handed to h2.

Expected: ping #1 succeeds. The transport notices the GOAWAY, reconnects, and sends the request on the new connection, as `HttpClient` does.

## Where in the source (0.9.0, paths relative to the crate root)

<details>
<summary>Source trace (connectrpc 0.9.0, hyper 1.11.0)</summary>

`src/client/http2.rs`: `SharedHttp2Connection::send` (1031-1047) goes through the tower `Buffer` to `Reconnect::poll_ready` (1322-1373) and `Reconnect::call` (1375-1396), and then to the `SendRequest` wrapper (1059-1076), whose `call` is `let fut = self.inner.send_request(req)` (1073). That is hyper's plain `send_request` with no retry; `try_send_request` is not used anywhere in the crate. The error is then mapped to `ConnectError::unavailable_from_transport("h2 send failed", e)` (1044).

`Reconnect` learns that a connection is gone only from `SendRequest::poll_ready` (1068-1070, handled in the `Connected` arm at 1361-1369, which logs "h2 connection lost; reconnecting" at 1367). hyper 1.11's `http2::SendRequest::poll_ready` checks `is_closed()` and nothing else (hyper `src/client/conn/http2.rs:97-103`). After a graceful GOAWAY with a stream still open, the connection task keeps running, so `is_closed()` stays false and the request is queued. hyper's h2 dispatcher, whose first act on waking is `h2_tx.poll_ready()`, sees `GoAway(NO_ERROR)` and returns `Dispatched::Shutdown` without dequeuing the request (hyper `src/proto/h2/client.rs:678-690`). Dropping the queue drops the envelope, whose destructor completes the caller with `Canceled("connection closed")` and hands the intact request back (hyper `src/client/dispatch.rs:219-227`). `send_request` discards that handed-back request; `try_send_request` (hyper `src/client/conn/http2.rs:181`) returns it to the caller.

`Reconnect` and the `SendRequest` wrapper are unchanged between 0.8.1 and 0.9.0 (0.8.1: 1021-1038, `send_request` at 1035).

</details>

## Suggested fix

Call `try_send_request` in the `SendRequest` wrapper, and when hyper hands the request back, reconnect and send it once more. This is what hyper-util's legacy client does (`retry_canceled_requests`, hyper-util 0.1.20 `src/client/legacy/client.rs:245-260`), and scenario B shows that behaviour working against the same server. It is safe for non-idempotent calls because hyper only hands back a request that never reached the wire. A local build of 0.9.0 with `SendRequest::call` switched to `try_send_request` reported the request as handed back for exactly this failure, so the resend can be implemented at that layer.

## Scope

Verified by running the program above on 0.9.0, and on 0.8.1 with the same scenarios in an equivalent program (0.8.1's `call_unary` / `call_server_stream` take `(service, method)` where 0.9.0 takes a `Spec`): the failure, its precondition, and that `HttpClient` and an idle `Http2Connection` are unaffected. The causal trace is from reading the connectrpc 0.9.0, hyper 1.11.0 and hyper-util 0.1.20 sources. Inferred from the dispatcher loop and not run: every request already queued on the connection when the dispatcher wakes is cancelled together, so a burst of requests routed to one connection can lose more than one to a single GOAWAY.

iainmcgin:
`SharedHttp2Connection` queued every request to a `tower::buffer` worker task. In the small-unary gRPC benchmark at c=1 the shared handle is measurably slower per call than `HttpClient` against the same server, and the gap closes when the client runtime has a single worker thread, which points at the cross-task wakeup rather than CPU work (bare-metal before/after numbers in a comment below). The worker also established readiness ahead of the next request, which is the mechanism behind #285: after a graceful GOAWAY with a stream still open, the next request was handed to the already-retired connection and failed unsent.

hyper's h2 `SendRequest` is itself a cloneable handle onto the connection task, so the shared handle now keeps the live `SendRequest` beside the reconnect state machine. Each request clones the handle and sends on it directly; only a request that finds no open connection locks the state machine, has a new connection established on a spawned task, and publishes the new handle. Requests queued behind an attempt share its outcome instead of dialling in turn, a request abandoned mid-handshake leaves the attempt to finish on its own task rather than half-done for the next request to inherit, and a request that hyper hands back unsent because it raced a closing connection is resent on a fresh handle, at most twice.

`Http2Connection::shared(bound)` keeps its signature and `bound` keeps its role: it was the worker's queue depth and is now the number of requests that may wait for the connection to be (re-)established at once, held as a semaphore slot that is released the moment the request is handed to hyper, so it never caps concurrent streams. As a `tower::Service` the handle keeps `Http2Connection`'s readiness contract: `poll_ready` drives re-establishment itself and reports a failed attempt as ready with the error returned by the next `call`, so it stays usable under `tower::balance`. `Clone` is now manual because a clone shares the connection but not the readiness a handle took in `poll_ready`.

Fixes #285.

### DIFF
--- a/.changes/unreleased/Changed-20260906-170500.yaml
+++ b/.changes/unreleased/Changed-20260906-170500.yaml
@@ -1,0 +1,13 @@
+kind: Changed
+body: |-
+    **`SharedHttp2Connection` sends directly on the hyper connection handle;
+    `bound` now caps in-flight requests.** Each request clones the live hyper
+    request handle and sends on it, taking the reconnect state machine's lock
+    only when there is no open connection, which removes the `tower::buffer`
+    worker task and its per-call hop and wakeups. `Http2Connection::shared(bound)`
+    keeps its signature: `bound` is now the number of requests that may be in
+    flight (sent and awaiting response headers) rather than the depth of the
+    worker's queue. It must still be non-zero. `shared` no longer has to be
+    called inside a Tokio runtime, and the workspace no longer enables tower's
+    `buffer` feature.
+time: 2026-09-06T17:05:00.000000000+00:00
--- a/.changes/unreleased/Changed-20260906-170500.yaml
+++ b/.changes/unreleased/Changed-20260906-170500.yaml
@@ -1,13 +1,18 @@
 kind: Changed
 body: |-
-    **`SharedHttp2Connection` sends directly on the hyper connection handle;
-    `bound` now caps in-flight requests.** Each request clones the live hyper
-    request handle and sends on it, taking the reconnect state machine's lock
-    only when there is no open connection, which removes the `tower::buffer`
-    worker task and its per-call hop and wakeups. `Http2Connection::shared(bound)`
-    keeps its signature: `bound` is now the number of requests that may be in
-    flight (sent and awaiting response headers) rather than the depth of the
-    worker's queue. It must still be non-zero. `shared` no longer has to be
-    called inside a Tokio runtime, and the workspace no longer enables tower's
-    `buffer` feature.
+    **`SharedHttp2Connection` sends directly on the hyper connection handle
+    instead of through a `tower::buffer` worker.** Each request clones the live
+    hyper request handle and sends on it, taking the reconnect state machine's
+    lock only when there is no open connection, which removes the worker task
+    and its per-call hop and wakeups. `Http2Connection::shared(bound)` keeps its
+    signature and `bound` keeps its role: it is the number of requests that may
+    wait for the connection to be (re-)established at once (previously the
+    worker's queue depth), never a cap on concurrent streams. It must still be
+    non-zero. As a `tower::Service` the handle keeps `Http2Connection`'s
+    readiness contract (`poll_ready` drives re-establishment and is pending
+    meanwhile; a failed attempt is ready with the error returned by the next
+    `call`), and `call` without a successful `poll_ready` returns an `internal`
+    error where the buffer used to panic. `shared` no longer has to be called
+    inside a Tokio runtime, and `Debug` output now includes the endpoint URI.
+    No downstream action is needed.
 time: 2026-09-06T17:05:00.000000000+00:00
--- a/.changes/unreleased/Changed-20260908-160000.yaml
+++ b/.changes/unreleased/Changed-20260908-160000.yaml
@@ -1,0 +1,9 @@
+kind: Changed
+body: |-
+    **Connect failures through `SharedHttp2Connection` keep their own error
+    code.** Failures the transport classified itself (such as the
+    `invalid_argument` scheme check or the establishment timeout) used to reach
+    the caller re-wrapped as `unavailable: h2 send failed: …`; they now arrive
+    with their original code and message. Unclassified connect failures read
+    `h2 connect failed: …` rather than `h2 send failed: …`.
+time: 2026-09-08T16:00:00.000000000+00:00
--- a/.changes/unreleased/Fixed-20260906-172500.yaml
+++ b/.changes/unreleased/Fixed-20260906-172500.yaml
@@ -1,0 +1,14 @@
+kind: Fixed
+body: |-
+    `SharedHttp2Connection`: the first request after a graceful GOAWAY arrived
+    while another stream was still open failed with `h2 send failed: operation
+    was canceled` without reaching the wire (#285). Readiness is now checked
+    when the request is made rather than ahead of time by a worker, so that
+    request goes out on a new connection, and a request that races a closing
+    connection is resent. Two related fixes: a request abandoned while it
+    was establishing the connection (for example at its deadline) no longer
+    leaves a half-finished attempt and its establishment timeout to the next
+    request (the attempt now runs to completion on its own task), and connect failures that the transport classified
+    itself (such as the `invalid_argument` scheme check) keep their code
+    instead of being re-wrapped as `unavailable`.
+time: 2026-09-06T17:25:00.000000000+00:00
--- a/.changes/unreleased/Fixed-20260906-172500.yaml
+++ b/.changes/unreleased/Fixed-20260906-172500.yaml
@@ -1,14 +1,15 @@
 kind: Fixed
 body: |-
-    `SharedHttp2Connection`: the first request after a graceful GOAWAY arrived
-    while another stream was still open failed with `h2 send failed: operation
-    was canceled` without reaching the wire (#285). Readiness is now checked
-    when the request is made rather than ahead of time by a worker, so that
-    request goes out on a new connection, and a request that races a closing
-    connection is resent. Two related fixes: a request abandoned while it
-    was establishing the connection (for example at its deadline) no longer
-    leaves a half-finished attempt and its establishment timeout to the next
-    request (the attempt now runs to completion on its own task), and connect failures that the transport classified
-    itself (such as the `invalid_argument` scheme check) keep their code
-    instead of being re-wrapped as `unavailable`.
+    **`SharedHttp2Connection` no longer fails the first request after a graceful
+    GOAWAY** ([#285]). A request made while another stream was still open on
+    the retiring connection failed with `h2 send failed: operation was
+    canceled` without reaching the wire. Readiness is now checked when the
+    request is made rather than ahead of time by a worker, so that request goes
+    out on a new connection, and a request that a closing connection hands back
+    before it reached the wire is resent on a fresh one. Relatedly, a request
+    abandoned while it was establishing the connection (for example at its
+    deadline) no longer leaves a half-finished attempt and its establishment
+    timeout to the next request: the attempt runs to completion on its own task.
+
+    [#285]: https://github.com/connectrpc/connect-rust/issues/285
 time: 2026-09-06T17:25:00.000000000+00:00
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ http-body-util = "0.1"
 bytes = { version = "1.6", features = ["serde"] }
 
 # Tower service abstraction
-tower = { version = "0.5", features = ["util", "buffer"] }
+tower = { version = "0.5", features = ["util"] }
 tower-http = { version = "0.6", features = ["catch-panic"] }
 
 # Async runtime
--- a/connectrpc/Cargo.toml
+++ b/connectrpc/Cargo.toml
@@ -61,7 +61,8 @@ serde_json = { workspace = true }
 flate2 = { workspace = true, optional = true }
 zstd = { workspace = true, optional = true }
 
-# Envelope framing (`Encoder`/`Decoder` for the 5-byte message header)
+# Envelope framing (`Encoder`/`Decoder` for the 5-byte message header) and
+# `PollSemaphore` for the shared h2 client handle
 tokio-util = { workspace = true, features = ["codec"] }
 
 # Error handling
--- a/connectrpc/src/client/http2.rs
+++ b/connectrpc/src/client/http2.rs
@@ -50,7 +50,6 @@ type BoxError = Box<dyn std::error::Error + Send + Sync>;
 // TLS support types and helpers
 // ============================================================================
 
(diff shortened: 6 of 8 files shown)


iainmcgin:
[claude code] Bare-metal A/B (AWS c7i.metal-24xl, Xeon Platinum 8488C, turbo off; connectrpc-rs `bench_server` from `main` pinned to 8 cores, load generator on 16 others, loopback). The client is a closed-loop gRPC small-unary driver over one `Http2Connection::shared(1024)` handle, built from `main` and from this branch (63fbc44), 10 s per cell, three rounds alternating the two clients:

| requests in flight | main | this branch | delta |
|---:|---:|---:|---:|
| 1 | 70.5 µs/req | 68.6 µs/req | −2.7% |
| 16 | 38.4k req/s | 38.6k req/s | +0.6% (noise) |
| 64 | 39.3k req/s | 39.5k req/s | +0.6% (noise) |

For reference, `HttpClient` against the same server measures 69.1 µs at one request in flight, so the shared handle goes from 2% slower than the pooled client to level with it. The same driver against the 64-byte echo server: 17.2k → 17.7k req/s at 1 in flight (+2.8%), within noise at 16 and 64. Round-to-round spread was under 1% at c=1 and about ±1.5% at c=16/64. An earlier revision of this branch (5114f72, before the queue-slot rework) measured the same within noise.

rpb-ant (approved):

PR #304, opened 2026-09-14
server: `Server::with_connection_extensions` — populate `ConnectionInfo` once per connection

rpb-ant:
Per-connection facts, such as an identity parsed from the client certificate, are computed once after the handshake instead of on every request:

```rust
connectrpc::axum::serve_tls(listener, app, tls) // or Server / BoundServer
    .with_connection_extensions(|conn, ext| {
        ext.insert(PeerIdentity::from_connection(conn)); // cloned into every request
    })
```

- New on `Server`, `BoundServer` and `connectrpc::axum::Serve`: `with_connection_extensions(impl Fn(&ConnectionInfo, &mut http::Extensions) + Send + Sync + 'static)`. The function reads the connection and fills an empty map, whose entries join the connection's extensions. It cannot remove an entry or change the peer: `PeerAddr` / `PeerCerts` still come only from the `ConnectionInfo` peer fields. It is synchronous and cannot refuse a connection; a custom loop around `Server::serve_connection` does that.
- It runs once per connection, on the connection's task, after the TLS handshake and before the first request, both in the shared accept loop and inside `Server::serve_connection`'s future. The free `server::serve_connection` does not run it. A panic in it is logged and drops only that connection, which `Server::serve_connection` reports as `CloseReason::Error`; under `panic = "abort"` it aborts the process.
- The function is stored in a newtype that asserts `UnwindSafe` / `RefUnwindSafe`, so `Server`, `BoundServer` and `axum::Serve` keep their auto traits.


iainmcgin (approved):
[claude code] Approving under the maintainers' agreement to review and merge each other's PRs. Rebased onto main after #308 merged; the added fix commits went through code, API, doc and deslop review.

Issue #250, opened 2026-07-24
Server and BoundServer duplicate their whole config surface because config cannot cross between them


iainmcgin:
## Summary

`Server` and `BoundServer` each declare the same nine connection-level fields and fourteen identical `with_*` setters, and the default block is written out in full four times. Adding one connection knob currently costs around 35 lines across roughly 20 hunks.

The usual reading of this is "deduplicate the builders". The more useful reading is that the duplication exists to work around a missing transition.

## The missing transition

`Server::bind` and `Server::from_listener` take **no `self`**:

```rust
// connectrpc/src/server.rs
pub fn from_listener(listener: TcpListener) -> BoundServer { ... }
pub async fn bind(addr: impl tokio::net::ToSocketAddrs)
    -> Result<BoundServer, Box<dyn std::error::Error + Send + Sync>> { ... }
```

They are associated functions that construct a fresh `BoundServer` with default settings. A configured `Server`'s only exit is `serve(self, addr)`. And `serve_with_graceful_shutdown` exists only on `BoundServer`.

So a user who wants graceful shutdown, or who needs `local_addr()` to bind port 0 in a test, has to abandon `Server` and re-apply every knob to `BoundServer`. That is why both types need the full setter surface.

There is a second cliff behind the first: the service-level knobs — `with_limits`, `with_compression`, `with_compression_policy`, `with_interceptor`, `with_deadline_policy` — exist only on `Server`. Taking the `BoundServer` route to get graceful shutdown silently drops the fluent path to those, and the user has to hand-build `ConnectRpcService::new(router).with_limits(..)` and call `serve_with_service`. Nothing in the signatures signals it and the docs on `bind` do not mention it.

## Evidence of the cost

Seven commits in about two months each added one connection knob: #165 (max connection age), #184 (max concurrent streams), #185 (max requests per connection), #186 (adaptive flow control), #188 (HTTP/2 keepalive), #189 (HTTP/1.1 header read timeout), #190 (max connection idle). #184's title — "add with_max_concurrent_streams to Server and BoundServer" — states the shape directly. #123 exists purely to re-sync a duplicated surface.

Two smaller symptoms of the same cause:

- `serve_with_listener` carries nine positional parameters under `#[allow(clippy::too_many_arguments)]`, one of which is behind `#[cfg(feature = "server-tls")]` — so the function's arity changes with features, and all three call sites must repeat the `cfg`.
[...]

PR #191, opened 2026-06-21
docs: document raw hyper transport config

Jaksenc:
## Problem

The hosting guide tells users to switch to axum when the built-in `Server`
does not fit, but `axum::serve` is not the escape hatch for transport-level
HTTP/2 settings. Users who need knobs such as flow-control windows, frame
size, HPACK table size, or exact keepalive behavior need to drive
`ConnectRpcService` with hyper directly.

## Change

- Narrow the standalone-server guidance so axum is recommended for routing,
  health checks, ordinary HTTP endpoints, and request-level Tower middleware.
- Add an "Advanced transport configuration" section showing the
  `ConnectRpcService` + `TowerToHyperService` + `hyper_util` auto-builder
  pattern.
- Cross-link from the `server` module docs so readers of `Server` can discover
  the raw-hyper path for unexposed transport/HTTP/2 knobs.

This keeps `Server`'s API focused while documenting the lower-level hosting
path that already exists through the Tower service implementation.

## Validation

- `rg -n 'Advanced transport configuration|TowerToHyperService|serve_connection|adaptive_window|max_concurrent_streams' docs/guide.md connectrpc/src/server.rs`
- `rg -n 'switching to the axum path|axum.*transport|axum::serve.*HTTP/2|axum::serve.*transport' docs/guide.md README.md connectrpc/src`
- `git diff --check`
- `cargo +nightly-2026-02-27 fmt --all -- --check`
- `RUSTDOCFLAGS="-Dwarnings" cargo doc -p connectrpc --all-features --no-deps`
- `cargo check -p connectrpc --all-features --offline`
- `cargo check -p connectrpc --no-default-features --features server --offline`

## Risk

The new example is intentionally a small hosting pattern, not a full production
accept loop. It calls out that raw hyper users must insert `PeerAddr` /
`PeerCerts` themselves if handlers rely on `ctx.peer_addr()` or
`ctx.peer_certs()`.

## Out of scope

- No new `Server` builder APIs.
- No axum behavior changes.
- No generated code, dependency, or lockfile changes.

## Issue linkage

Closes #179

PR #305, opened 2026-09-15
RFC: server internals as acceptor + driver + loop (opt-in restructure; follows #306–#308, #304)

### LINKED ISSUE #253: Test modules run several times the size CONTRIBUTING allows, with the harness copied per test
## Summary

CONTRIBUTING says test modules beyond roughly 500 lines should be split into focused sub-modules. Nine files exceed that; three exceed it fourfold. There is also no shared test-support crate, so the same server spin-up and tear-down is written out dozens of times.

| File | Test-module lines |
|---|---|
| `connectrpc/src/client/mod.rs` | 3,329 |
| `connectrpc-codegen/src/codegen.rs` | 2,419 |
| `connectrpc/src/server.rs` | 2,198 |
| `tests/streaming/src/lib.rs` | 2,002 |
| `connectrpc/src/service.rs` | 2,020 |
| `connectrpc/src/interceptor.rs` | 1,079 |
| `connectrpc/src/compression.rs` | 992 |
| `connectrpc/src/response.rs` | 699 |
| `connectrpc-build/src/lib.rs` | 694 |

## The duplication, ranked by what it risks

**1. The conformance client's configuration prologue, four times, byte-identical.** In `conformance/src/bin/client.rs`, each of `do_unary_call`, `do_server_stream_call`, `do_client_stream_call` and `do_bidi_stream_call` opens with the same ~45 lines building the base URI, the `ClientConfig`, and the `CallOptions` from the runner's request. This is the one that matters most, because the code is semantically load-bearing: it translates the conformance runner's `ClientCompatRequest` into client configuration, so a fix applied to one arm and not the other three silently changes which conformance tests pass. The only real variation is the compression registry, and both variants already carry comments explaining why they differ.

**2. `server.rs`'s lifecycle boilerplate.** The spawn-with-graceful-shutdown prologue appears 22 times and its tear-down epilogue 24 times. Two tests are 39 and 35 lines that differ only in one builder call and one assertion, duplicated five-line comment included. The file already has a small helper block, so the instinct is there; it just stopped short of the lifecycle. Separately, eight tests bind a real TCP socket purely to assert on a `BoundServer` config field they never serve on.

**3. `benches/rpc`'s load drivers.** `echo_bench.rs`, `log_bench.rs` and `fortune_bench.rs` each re-implement `ServerProcess` with its `Drop`, a `build_*_server` function differing only in a binary name, a `BenchResult` struct, and a ~100-line closed-loop driver. It has already drifted — one copy of the driver silently dropped its explanatory comment and diverged on a sampling constant. All of this runs in the load-generator process, out of band from the server being measured, so factoring it cannot perturb a number.

## Split axes

These follow how the tests already cluster, and match the structure of the production half of each file:

- **`server.rs`** — by configuration surface: builder-config assertions, graceful shutdown, header-read timeout, HTTP/2 knobs, connection lifecycle (age/idle/count), TLS peer identity, shared helpers.
- **`client/mod.rs`** — by protocol then concern: bidi lifecycle, Connect stream decoding, gRPC/gRPC-Web decoding, transport-error propagation, client-streaming send/drop, timeout formatting, header construction, body helpers, config merge. The timeout-format and body-helper clusters are pure-function tests with no client involved and could move next to what they test.
- **`codegen.rs`** — by codegen feature: fixtures, view/alias emission, file-per-package, type-path resolution, keyword escaping, client-feature gating, `encodable_impls`, whole-output invariants. The plugin-option tests are currently scattered across seven separate places; collecting those alone is the cheapest useful move.
- **`tests/streaming`** — the cleanest one. It is the workspace's designated end-to-end consumer of `connectrpc-build`, it has no `tests/` directory, and 36 tests plus the service impls sit in `src/lib.rs`. Promoting the shared harness into a `pub mod testing` and moving the tests to `tests/` also gets process-level parallelism for free.

## Deliberately excluded

The example crates' `start_server` helpers should stay duplicated. Each is eight to fourteen lines and each differs in exactly the thing its crate exists to demonstrate. A shared helper covering all six would have to be generic over the client type, the registration call and the layer stack, and would hide the API being demonstrated to save 60–80 lines.

The bench *server* binaries should also stay separate. `log_server` and `log_server_noutf8` differ in their handler bodies precisely where the measurement is, and sharing a handler behind a trait would put a call boundary in the measured path.

The conformance runner's flat one-function-per-RPC-kind shape is correct and should be kept — a conformance driver's job is to be readable against the spec. The fix there is the duplicated prologue, not the shape.

## Suggested order

Each of these is independently landable under the 250-line-net guideline: the conformance prologue first (guarded by 6,800 conformance cases), then the bench harness, then a `TestServer` helper in `server.rs`, then the module splits one file at a time.

rpb-ant:
This retitles #305. It is **not** required for any of the API in A–D (#306 `ConnectionInfo` + `Server::serve_connection`, #307 `ConnectionConfig`, #308 free `serve_connection` + `axum::serve`, #304 `with_connection_extensions`) — those four land the whole adopter-facing surface on today's `server.rs` / `axum.rs`, each under the 250-line rule by net non-test lines (B's gross churn is ~2k because setter docs move; it is mechanical), with no file moves. What is left in this branch is internal structure, offered because it addresses #250 (items 2–3 and the Notes line) and #253 (the `server.rs` row) directly. If you want it, I will rebase it onto D and send it as 3–4 mostly-move PRs; if not, nothing in A–D depends on it and it can be closed.

The branch as pushed (`rpb/server-layers`, 10 commits on `main`) **predates A–D** and still carries the earlier, larger public surface (public `Acceptor` family, `Server::service`, `accept_config` accessors). Read it for shape, not for API: rebased, every one of those becomes `pub(crate)` and the public items are exactly A–D's.

## What A–D already bank (so this RFC does not claim them)

- One connection driver behind `Server`, `BoundServer`, `axum::serve`, `axum::serve_tls` and custom loops (C), returning `ConnectionClosed` (A).
- One crate-private accept loop shared by all four entry points (C) — `serve_tls` can no longer silently lack a setting.
- One `ConnectionConfig` value instead of nine fields × two builders × four default blocks × positional plumbing (B): a new knob is field + default + setter + accessor + apply = 5 hunks, versus the measured 11 (HTTP/2 knob) to 22 (positional knob; the fourth default block really does fail `cargo check`) on 0.9.0, and it cannot skip `axum.rs` as 4 of the last 5 knob PRs did.

## What the restructure adds, measured on the branch against `main` @ 758bd3a

| | main | branch |
|---|---|---|
| `server.rs` | 4 330 lines (2 132 non-test) in one file | `server/{mod 134, config 838, peer 259, acceptor 369, connection 621, accept_loop 125, standalone 784}.rs`; largest non-test file 707 |
| accept loops / TLS handshake blocks | 2 / 2 | 1 / 1 (`accept_loop::run`, 72 lines; `Accepted::handshake`, 28) — already 1 / 1 after C; what the branch adds is the loop and the accept+handshake step as their own small units (`accept_loop.rs`, `acceptor.rs`) instead of one 130-line function |
| longest function | `serve_with_listener` 174, `ConnectionLifecycle::poll` 115 | `serve_connection` 126, `accept_loop::run` 72 |
| connection lifecycle | hand-rolled `Future`: struct + `IdleTracker` + 3-state enum + `poll` with 10 poll sites, 210 lines | two `biased` `select!` phases + `retirement()` + `quiet_for`, 99 lines (last commit, −177/+90, independently droppable) |
| `#[allow(clippy::too_many_arguments)]` / cfg-dependent arity | 2 / yes | 0 / no |
| tests | one 2 198-line `mod tests` (59 tests), 0 in-memory, 41 sockets | `server/tests/{acceptor,builders,connection,header_read_timeout,http2,max_connection_age,max_connection_idle,max_requests,peer,shutdown}.rs` by topic (#253's own axes) + unit tests beside the code; a direct driver suite of 8 tests over `tokio::io::duplex` (max-age / idle / request-count retirement, drain, panic isolation, extension stamping) with no listener |
| functional diff (driver + loop + axum as one change) | — | **net −405 non-test lines**, because the second loop and handshake die; commits 1–2 are a 100 % rename and a pure test move (+23 net) |

Costs, also measured: 9 `pub(crate)` seams between files that were one; `cfg(feature = "server-tls")` sites 37 → 41 spread over 5 files instead of 2; the layer table has to be kept true in module docs, crate docs and the guide; +4 CI steps (already in C).

## Proposed shape if accepted (rebased onto D)

1. **moves** — `server.rs` → `server/mod.rs`; test module → `server/tests/*.rs` by topic. Pure moves, reviewable with `--color-moved`; closes #253's `server.rs` row on its own.
2. **acceptor + loop** — crate-private `Acceptor` / `Accepted::handshake` (TCP accept, nodelay, transient-error retry with a 1 s back-off on `EMFILE`/`ENFILE`, TLS handshake + timeout + cert capture in one place) and `accept_loop::run` in their own files; `standalone.rs` becomes the two builders + `run`. No public items.
3. **async lifecycle** — the `poll` state machine as an async body. Droppable.
4. **example + docs** — `examples/custom-accept-loop` (admission and runtime placement by client-certificate identity over the public `serve_connection`, handshake inline as in C's guide section, with an e2e test) and the module-doc layer table.

Keeping `Acceptor` / `Accepted` / `ServerIo` / `HandshakeError` / `AcceptConfig` crate-private until someone asks is deliberate: promoting them later is ~24 visibility tokens and re-exports; retracting them would be a breaking release, and #250's `self`-taking transition may want to reshape `AcceptConfig` first.

Evidence from the branch as it stands: per-commit fmt / clippy (all feature shapes) / rustdoc / MSRV green; 697 lib tests; conformance server mode 3600/3600 (all protocols) and 2396/2396 (Connect + TLS); `cargo semver-checks` vs v0.9.0 clean three ways; `echo_bench --quick` within noise of `main`.

[#49]: https://github.com/connectrpc/connect-rust/issues/49
[#191]: https://github.com/connectrpc/connect-rust/issues/191
[#250]: https://github.com/connectrpc/connect-rust/issues/250
[#253]: https://github.com/connectrpc/connect-rust/issues/253


rpb-ant:
Closing: the parts of this with adopter demand landed as #306, #307, #308 and #304 (one `ConnectionConfig`, `ConnectionInfo`, a public service-generic `serve_connection`, and `Server`/`BoundServer`/`axum::serve` sharing one accept loop and driver).

What remains of the layering proposal is internal and behaviour-neutral: splitting `server.rs` into `server/{config,peer,connection,accept_loop,…}.rs` (the #253 row), splitting the test module by topic with a socket-free driver suite, extracting the two inline TLS-handshake blocks into a crate-private acceptor, and expressing the connection lifecycle as an `async` body instead of the hand-rolled `poll` state machine. The branch here predates #306–#308 and their follow-up fixes, so it would be re-cut from `main` as move-only commits rather than rebased. Happy to do that whenever #253/#250 want it; leaving it as an offer rather than an open diff.

PR #306, opened 2026-09-16
server: public `ConnectionInfo`; `Server::serve_connection` returning `ConnectionClosed`

rpb-ant:
Before this, owning the accept step meant raw hyper, which drops every timeout, retirement and drain guarantee. After:

```rust
let mut info = ConnectionInfo::new().with_peer_addr(peer).with_peer_certs(chain);
info.extensions_mut().insert(tenant); // cloned into every request
let closed = server.serve_connection(stream, info, drain.clone().wait()).await;
metrics.record(closed.reason()); // Closed | Shutdown | MaxAge | Idle | MaxRequests | Error
```

- `#[non_exhaustive] ConnectionInfo`, formerly the private `PeerInfo`, also at the crate root. `PeerAddr` / `PeerCerts` on requests always come from its peer fields, never from the extensions.
- `Server::serve_connection(&self, io, info, shutdown) -> impl Future<Output = ConnectionClosed> + Send + 'static` serves one already-accepted stream with the server's service and connection settings, on whichever runtime polls the future. `Server::with_tls` does not apply here: the caller completes any TLS handshake first.
- `ConnectionClosed::reason() -> CloseReason` (`#[non_exhaustive]`). Once shutdown or a retirement trigger tells a connection to wind down, that reason is final: a peer that errors mid-drain, or that sent nothing before shutdown, reports the wind-down reason, not `Error`.

First of four stacked PRs: #306 → #307 → #308 → #304. Part of #191 and #49.


iainmcgin (approved):
[claude code] Approving under the maintainers' agreement to review and merge each other's PRs. The added fix commits went through code, API, doc and deslop review; CI, server conformance and the MSRV check pass.

Issue #247, opened 2026-07-24
Limits setters break the crate's with_ convention and duplicate its public fields


iainmcgin:
## Summary

`Limits` is the only runtime-config type in the crate whose setters are bare verbs rather than `with_`-prefixed, and it is also the only one that exposes public fields alongside those setters — with the field and the method sharing an identifier.

```rust
// connectrpc/src/service.rs
#[non_exhaustive]
pub struct Limits {
    pub max_request_body_size: usize,
    pub max_message_size: usize,
    pub element_memory_limit: usize,
}

impl Limits {
    pub fn max_request_body_size(mut self, size: usize) -> Self { ... }
    pub fn max_message_size(mut self, size: usize) -> Self { ... }
    pub fn element_memory_limit(mut self, bytes: usize) -> Self { ... }
}
```

Every sibling config type uses `with_`: `ConnectRpcService::with_limits` / `with_compression`, `ClientConfig::with_protocol` / `with_default_timeout`, `Server::with_*`, `RequestContext::with_*`. There are 81 `with_`-prefixed setters in the crate and three that are not, all three on this type.

## Why it matters

A user who has learned `with_` on the service and the client reaches for `limits.with_max_message_size(..)`, gets method-not-found, and has to read the source to discover the bare name. That is a small cost, but it is paid by every user exactly once at the point where they are hardening a server, and it is avoidable.

The overlap with the fields is the sharper problem. Because the fields are `pub` and the methods share their names, there are two ways to set each value; and because the struct is `#[non_exhaustive]`, struct-update syntax does not work anyway, so the public fields buy construction flexibility that the attribute has already removed. What they do buy is field *reads*, which accessors would cover.

`ClientConfig` is the shape to copy: private fields, `with_` setters, plain accessors, `#[non_exhaustive]`, and a doc note about the struct-update limitation.

## Proposal

1. Rename the three setters to `with_max_request_body_size`, `with_max_message_size`, `with_element_memory_limit`.
2. Make the three fields private and add accessors returning `usize`.

Both are breaking, which is why this belongs in the 0.9.0 window rather than after it.

Item 1 alone is the higher-value half; item 2 is worth taking at the same time so the type is only broken once, but is separable if there is disagreement about it.

## Notes

Related to #173 (rejecting illegal values in configuration setters) in that both are about making the setter surface consistent, but the changes do not overlap — that issue is about validation behaviour, this one is about naming and field visibility.

iainmcgin:
[...]

PR #307, opened 2026-09-16
server: `ConnectionConfig` — one per-connection settings value held by `Server` and `BoundServer`

rpb-ant:
Implements #250 items 2–3, modelled on `ClientConfig` from #247. Most of the `server.rs` diff moves the long setter docs from `BoundServer` onto `ConnectionConfig`; `git diff --color-moved=dimmed-zebra` separates the moves from the edits.

Before this, nine per-connection fields were duplicated on `Server` and `BoundServer`, with four hand-synchronised default blocks and a nine-argument `serve_with_listener`. After, each builder holds one value, and the same value configures either:

```rust
let config = ConnectionConfig::new()
    .with_max_connection_age(Duration::from_secs(600))
    .with_http2_max_header_list_size(64 << 10);
Server::new(router).with_connection_config(config.clone()).serve(addr).await?;
// or: Server::bind(addr).await?.with_connection_config(config).serve(router).await?;
```

- Public `ConnectionConfig`, also at the crate root (`#[non_exhaustive]`; derives `Clone`, `Debug`, `PartialEq`, `Eq`), holds every setting the builders apply to a connection. The builders' existing `with_*` setters keep their signatures and edit that value; `with_connection_config` / `connection_config()` replace or return it whole.
- New setting `with_http2_max_header_list_size(u32)`, the advertised `SETTINGS_MAX_HEADER_LIST_SIZE` (hyper's 16 KiB default is too small for some bearer tokens), with shorthands on `Server` and `BoundServer`. It panics on zero. A request whose header list reaches the limit gets `431`. With h2 0.4.15 and later, a connection-closing GOAWAY follows a header list past four times the limit, or a header block spread over more CONTINUATION frames than h2 allows, which can happen well under four times the limit.
- The TLS settings move into a crate-private `AcceptConfig`, so the accept loop is `serve_with_listener(listener, service, accept, config, shutdown)` in every feature set.


iainmcgin (approved):
[claude code] Approving under the maintainers' agreement to review and merge each other's PRs. Rebased onto main after #306 merged; the added fix commits went through code, API, doc and deslop review.

PR #308, opened 2026-09-16
server: free `serve_connection` over any HTTP service; `axum::serve` / `serve_tls` on the shared loop and driver

rpb-ant:
The connection driver becomes a public free function over any tower HTTP service. `connectrpc::axum::serve_tls`, which had its own copy, and the new `connectrpc::axum::serve` now run on it and on `Server`'s accept loop.

```rust
connectrpc::axum::serve(listener, app) // plaintext, new
    .with_connection_config(ConnectionConfig::new().with_max_connection_age(age))
    .with_graceful_shutdown(signal)
    .await?;
// a custom loop, any tower HTTP service:
let closed = server::serve_connection(stream, info, app.clone(), config.clone(), drain).await;
```

- `server::serve_connection(io, info, service, config, shutdown)` binds nothing to a runtime until its future is first polled, so a loop chooses where a connection runs by where it spawns the future. A tokio socket stays on the I/O driver that registered it; the guide's "Custom accept loops" section moves it with `into_std` / `TcpStream::from_std`, which does not work once the stream is wrapped in TLS.
- `connectrpc::axum` now needs `axum` + `server`; `serve_tls` still needs `server-tls`. `ServeTls` is renamed `Serve`, with a `#[deprecated(since = "0.10.0")]` alias, and `Serve::with_connection_config` is new.
- On `serve_tls`, a handler that panics before responding now gets a `500` with a Connect `internal` body instead of a dropped connection, the standalone server's connection defaults and `ConnectionConfig` apply, and the `Serve` future owns the connection tasks.
- On every path, a panic while a response body is produced is logged and resets the stream (`INTERNAL_ERROR`) or closes the HTTP/1.1 connection, instead of unwinding through the connection task. HTTP/2 extended CONNECT is advertised, and the accept loop pauses for up to a second after `EMFILE` / `ENFILE`.
- Fixes a bug also in 0.9.0: with `max_connection_idle` set, a request stopped counting as in flight once its response head was returned, so a live streaming response could be closed as idle. It now counts until hyper drops the body, so a client that stops reading keeps its connection until `max_connection_age` retires it.

Part of #191 and #49.


iainmcgin (approved):
[claude code] Approving under the maintainers' agreement to review and merge each other's PRs. Rebased onto main after #307 merged; the added fix commits went through code, API, doc and deslop review.

